PR #59, opened 2021-10-14
feature: Support PostgreSQL, MySQL and SQLite databases.

elliotcourant:
Adding support for bun and multiple databases.


elliotcourant:
- MySQL does not support `RETURNING` clauses. Which are used by almost every insert query to get back the generated unique id of the object being created.
- SQLite does not support AUTO_INCREMENT on anything other than an integer primary key. Big integers are not supported, and composite primary keys are not supported with auto-increment. 

---

At the moment I may stop work on supporting other databases for now.

elliotcourant:
I'm going to try to take another swing at this once the new-ui is complete. I won't be supporting MySQL, but I will try to support both PostgreSQL and SQLite.

I really want to be able to support SQLite because it's way easier to distribute for people.

PR #334, opened 2021-11-16
fix: Upgrading react-router to v6.

elliotcourant:
Upgrading everything to react-router v6. But will probably switch to
wouter in the future given how tedious react-router has become.

This also makes several other changes. It furthers our migration to
functional components and using react hooks. It also completely removes
react-hot-loader.

On the backend the authentication approach has been changed as well. The
`M-Token` cookie is now `HttpOnly` which does not allow the UI to see it
at all. This is a huge security improvement; but requires some changes
in how authentication is performed by the frontend.

Now the frontend will make a request to `/users/me` upon bootstrapping.
If the request succeeds, then the frontend assumes the token is present
(even if invisible to the UI code) and proceeds to bootstrap
authentication. If the request fails, the UI assumes that the user is
not authenticated and falls back to the login page.

On the backend, the cookie is read for each authenticated request. If
the cookie is not valid for a request. The response for that request
will remove the cookie.

Because the UI cannot directly interact with the cookie, an
`/authentication/logout` path has been added to force removal of the
authentication token.


package.json
```diff
@@ -123,12 +120,12 @@
     "@tailwindcss/postcss7-compat": "npm:@tailwindcss/postcss7-compat",
     "@testing-library/jest-dom": "5.15.0",
     "@testing-library/react": "12.1.2",
+    "@testing-library/react-hooks": "^7.0.2",
```
elliotcourant, line 123: Dependency should be pinned

pkg/controller/authentication.go
```diff
@@ -129,10 +130,21 @@ func (c *Controller) loginEndpoint(ctx iris.Context) {
...
+			Name:     "M-Token",
+			Value:    token,
+			Path:     "/",
+			Domain:   c.configuration.APIDomainName,
+			Expires:  time.Now().Add(7 * 24 * time.Hour),
+			MaxAge:   0,
+			Secure:   true,
+			HttpOnly: true,
+			SameSite: http.SameSiteStrictMode,
+		})
```
elliotcourant, line 143: This is duplicated in several spots. It should probably be lifted into its own function and allow for other parameters like expiration, same site and secure be configurable.

HttpOnly should be a general requirement, the UI does not _need_ to know the actual token. 

But other things like secure or domain name should definitely be configurable for self-hosted deployments.

pkg/controller/authentication.go
```diff
@@ -164,13 +175,53 @@ func (c *Controller) loginEndpoint(ctx iris.Context) {
...
+		Name:     "M-Token",
+		Value:    "",
+		Path:     "/",
+		Domain:   c.configuration.APIDomainName,
+		Expires:  time.Now(),
+		MaxAge:   0,
+		Secure:   true,
+		HttpOnly: true,
+		SameSite: http.SameSiteStrictMode,
+	})
```
elliotcourant, line 222: This should use the all-in-one method previously mentioned; and should just pass an empty token.

pkg/controller/main_test.go
```diff
@@ -147,13 +154,28 @@ func NewTestApplicationWithConfig(t *testing.T, configuration config.Configurati
...
+	loginRequest.Password = password
+	var token string
+	{
+		response := e.POST("/api/authentication/login").
+			WithJSON(loginRequest).
+			Expect()
+
+		response.Status(http.StatusOK)
+		response.Cookie("M-Token").Value().NotEmpty()
+		token = response.Cookie("M-Token").Value().Raw()
```
elliotcourant, line 172: There is a const for the token name, this const should be used instead.

----

Maybe even make the name a configuration variable?

pkg/controller/middleware.go
```diff
@@ -100,10 +100,8 @@ func (c *Controller) authenticateUser(ctx iris.Context) (err error) {
 		}()
 	}
 
-	if token = ctx.GetCookie(TokenName, iris.CookieSecure); token != "" {
+	if token = ctx.GetCookie(TokenName, iris.CookieSecure, iris.CookieHTTPOnly(true)); token != "" {
```
elliotcourant, line 103: This should also use the configuration, if the cookie not secure (someone is just using http because its locally hosted) then we can't only look for secure cookies.

pkg/controller/middleware.go
```diff
@@ -157,6 +155,17 @@ func (c *Controller) authenticateUser(ctx iris.Context) (err error) {
...
+			Name:     "M-Token",
+			Value:    "",
+			Path:     "/",
+			Domain:   c.configuration.APIDomainName,
+			Expires:  time.Now(),
+			MaxAge:   0,
+			Secure:   true,
+			HttpOnly: true,
+			SameSite: http.SameSiteStrictMode,
+		})
```
elliotcourant, line 168: Should call the same code as logout.

ui/components/BankAccounts/BankAccountSelector.tsx
```diff
@@ -16,166 +13,131 @@ import fetchSpending from 'shared/spending/actions/fetchSpending';
...
+    .toArray();
+
+  items.push(<Divider key="divider" className="w-96"/>);
+  items.push(bankAccountsViewButton);
+
+  return (
+    <Fragment>
+      { items }
+    </Fragment>
+  );
```
elliotcourant, line 84: This is really dumb, but leave a comment on why this is being done.

ui/components/Plaid/PlaidButton.tsx
```diff
@@ -41,8 +43,10 @@ const HookedPlaidButton = (props: HookedPropTypes) => {
     open();
   };
 
-  const newProps: HookedPropTypes = {
-    ...props,
+  const { useCache, plaidOnSuccess, plaidOnExit, plaidOnLoad, plaidOnEvent, token, ...buttonProps } = props;
```
elliotcourant, line 46: Leave a comment about this, this is dumb.

ui/components/Plaid/PlaidButton.tsx
```diff
@@ -88,8 +92,9 @@ export default class PlaidButton extends Component<PropTypes, State> {
 
   renderButton = (): React.ReactNode => {
     const disabled = this.state.loading || this.props.disabled || this.state.disabled;
+    const { useCache, plaidOnSuccess, plaidOnExit, plaidOnLoad, plaidOnEvent, ...buttonProps } = this.props;
```
elliotcourant, line 95: Leave a comment about this, this is dumb.

ui/index.tsx
```diff
@@ -9,7 +9,9 @@ import './styles/styles.css';
 import './styles/index.scss';
 import * as Sentry from '@sentry/react';
 
-axios.get('/api/sentry')
+axios.get('/api/sentry', {
+  timeout: 3000,
```
elliotcourant, line 13: Leave a comment about this.

ui/models/Login.ts
```diff
@@ -0,0 +1,6 @@
+export default class Login {
+  loginId: number;
+  email: string;
+  firstName: string;
+  lastName: string;
+};
```
elliotcourant, line 6: Add a constructor for this as well.

ui/models/User.ts
```diff
@@ -0,0 +1,16 @@
...
+  userId: number;
+  loginId: number;
+  accountId: number;
+  login: Login;
+  firstName: string;
+  lastName: string;
+
+  constructor(data?: Partial<User>) {
+    if (data) {
+      Object.assign(this, data);
```
elliotcourant, line 13: For the login field, make sure the constructor is called.

ui/shared/authentication/actions/signUp.ts
```diff
@@ -0,0 +1,36 @@
...
+  isActive: boolean;
+  message: string | null;
+  nextUrl: string | null;
+  requireVerification: boolean | null;
+  user: User | null;
+}
+
+export interface SignUpError {
+  error: string;
+}
```
elliotcourant, line 26: This isnt even really used. We can just get rid of it.

elliotcourant: It is kind of used 🤷

PR #409, opened 2021-11-30
feature: Adding forgot password support.

### LINKED ISSUE #92: feature: Add support for users to reset their password via a "Forgot Password" flow.
Users need to be able to reset their password if they lose/forget it.

- [x] Add a "Forgot Password" button to the login page; it should redirect to a view that prompts for an email.
- [x] Add a forgot password API endpoint; the endpoint should always return a successful status code if a _valid_ email is provided. Even if that email is not associated with any login.
- [x] If the email address provided is associated with a login; send that user an email with a link that will let them reset their password. (not sure yet how this link will work, **provide feedback if you have ideas**).
- [x] The link should expire after a short period of time.
- [x] The link should only be able to be used once.
- [x] Document the forgot password API endpoint.

elliotcourant:
Resolves #92

### DIFF
--- a/pkg/config/configuration.go
+++ b/pkg/config/configuration.go
@@ -25,6 +25,10 @@ const (
 )
 
 type Configuration struct {
+	// configFile is not an actual configuration variable, but is used to let usages know what file was loaded for the
+	// configuration.
+	configFile string `yaml:"-"`
+
 	// DEPRECATED: This is not used anymore. It serves no function at all.
 	Name string `yaml:"name"`
 	// DEPRECATED: This is not used anymore. Use Server.ListenPort instead.
@@ -48,6 +52,10 @@ type Configuration struct {
 	Sentry        Sentry     `yaml:"sentry"`
 	Stripe        Stripe     `yaml:"stripe"`
 	Vault         Vault      `yaml:"vault"`
+}
+
+func (c Configuration) GetConfigFileName() string {
+	return c.configFile
 }
 
 func (c Configuration) GetUIDomainName() string {
@@ -161,6 +169,9 @@ type ReCAPTCHA struct {
 	Version        int    `yaml:"version"` // Currently only version 2 is supported by the UI.
 	VerifyLogin    bool   `yaml:"verifyLogin"`
 	VerifyRegister bool   `yaml:"loginRegister"`
+	// VerifyPasswordReset determines whether or not the user will be required to verify that they are not a robot
+	// overlord.
+	VerifyPasswordReset bool `yaml:"verifyPasswordReset"`
 }
 
 func (r ReCAPTCHA) ShouldVerifyLogin() bool {
@@ -169,6 +180,10 @@ func (r ReCAPTCHA) ShouldVerifyLogin() bool {
 
 func (r ReCAPTCHA) ShouldVerifyRegistration() bool {
 	return r.Enabled && r.VerifyRegister
+}
+
+func (r ReCAPTCHA) ShouldVerifyPasswordReset() bool {
+	return r.Enabled && r.VerifyPasswordReset
 }
 
 type Plaid struct {
@@ -322,6 +337,8 @@ func LoadConfigurationEx(v *viper.Viper) Configuration {
 		panic(err)
 	}
 
+	config.configFile = v.ConfigFileUsed()
+
 	return config
 }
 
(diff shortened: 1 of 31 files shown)


ui/views/Authentication/ForgotPasswordView.tsx
```diff
@@ -0,0 +1,137 @@
...
+      }
+    }
+
+    return errors;
+  }
+
+  function submitForgotPassword(values: ForgotPasswordValues, helpers: FormikHelpers<ForgotPasswordValues>): Promise<void> {
+    helpers.setSubmitting(true);
+
+    return sendForgotPassword(values.email, verification);
```
elliotcourant, line 43: Add a path that sets submitting back to false.

ui/views/Authentication/ForgotPasswordView.tsx
```diff
@@ -0,0 +1,137 @@
...
+      <div className="absolute top-2.5 left-2.5">
+        <Button
+          className="pr-3"
+          component={ RouterLink }
+          to="/login"
+        >
+          <ArrowBack className="mr-1"/>
+          Back To Login
+        </Button>
+      </div>
```
elliotcourant, line 57: This could be lifted into a re-usable component.

ui/views/Authentication/ForgotPasswordView.tsx
```diff
@@ -0,0 +1,137 @@
...
+             handleSubmit,
+             isSubmitting,
+             submitForm,
+           }) => (
+          <form onSubmit={ handleSubmit } className="h-full overflow-y-auto pb-20">
+            <div className="flex items-center justify-center w-full h-full max-h-full">
+              <div className="w-full p-10 xl:w-3/12 lg:w-5/12 md:w-2/3 sm:w-10/12 max-w-screen-sm sm:p-0">
+                <div className="flex justify-center w-full mb-5">
+                  <img src={ Logo } className="w-1/3"/>
+                </div>
```
elliotcourant, line 78: This is also used in several spots in the authentication views. This could be lifted into its own component.

ui/views/Authentication/ForgotPasswordView.tsx
```diff
@@ -0,0 +1,137 @@
...
+                      onChange={ handleChange }
+                      value={ values.email }
+                      variant="outlined"
+                    />
+                  </div>
+                  <CaptchaMaybe
+                    show
+                    loading={ isSubmitting }
+                    onVerify={ setVerification }
+                  />
```
elliotcourant, line 109: This will show whenever the `ReCAPTCHA` key is present, but not necessarily when we require captcha for this action. Have "show" based on the config.

ui/views/Authentication/ForgotPasswordView.tsx
```diff
@@ -0,0 +1,137 @@
...
+                  <div className="w-full pt-2.5 mb-10">
+                    <Button
+                      className="w-full"
+                      color="primary"
+                      disabled={ isSubmitting || !values.email || (requireCaptcha && !verification) }
```
elliotcourant, line 114: have `requireCaptcha` be based on the config as well.

ui/views/Authentication/ResetPasswordView.tsx
```diff
@@ -0,0 +1,168 @@
...
+
+  useEffect(() => {
+    if (!token) {
+      navigate('/login');
+      enqueueSnackbar('You must get a password reset link to change your password.', {
+        variant: 'warning',
+        disableWindowBlurListener: true,
+      });
+    }
+  }, [token, enqueueSnackbar, navigate]);
```
elliotcourant, line 39: If the token is provided, is there an easy way we can clear it from the URL? Makes the url prettier but also prevents the user accidently navigating to it again through the back button after the token has been used.

elliotcourant: @th3fallen do you know how I could change the URL without re-rendering this boi?

ui/views/Authentication/ResetPasswordView.tsx
```diff
@@ -0,0 +1,168 @@
...
+
+  function validateInput(values: ResetPasswordValues): FormikErrors<ResetPasswordValues> {
+    let errors: FormikErrors<ResetPasswordValues> = {};
+
+    if (values.password) {
```
elliotcourant, line 44: Should add something to trim the space with the password. Leading/trailing spaces are trimmed by the API anyway.

pkg/config/configuration.go
```diff
@@ -25,6 +25,9 @@ const (
 )
 
 type Configuration struct {
+	// ConfigFile is not an actual configuration variable, but is used to let usages know what file was loaded for the
+	// configuration.
+	ConfigFile string `yaml:"-"`
```
elliotcourant, line 30: Make this a private field, and have it only accessible via a method to make it immutable.

PR #456, opened 2021-12-10
feature: Revamped background job processing.

elliotcourant:
This PR aims to add significantly improved background job processing. With three main goals.

### Engines

- [ ] Support an in-memory job queue that is ideal for single replica self-hosted deployments. **Will be addressed separately**
- [x] Support a Redis based job queue that is ideal for multi-replica self-hosted deployments.
- [ ] Support a RabbitMQ job queue that is ideal for production worthy high-availability deployments. **Will be addressed separately**

Some job queues have varying requirements for recurring job scheduling. Job scheduling is used to kick off other jobs or jobs that need to be run periodically regardless of user interaction. These are things like periodically requesting Transactions from Plaid, processing funding schedules, or cleaning up old data from the database.

### Scheduling

Some job engines can handle the scheduling for these tasks themselves, and the scheduling is done via a cron syntax in the configuration (TODO).

Other job engines require that these scheduled jobs are triggered externally, as there is not a (simple) way to make sure these jobs are only triggered once if there are many replicas of monetr running.

| Job Engine           | Internal | External |
|----------------------|----------|----------|
| In-Memory            | ✅        | ❌        |
| gocraft/work (Redis) | ✅        | ✅        |
| RabbitMQ             | ❌        | ✅        |

---

These various job queue backends should be nearly interchangeable. The person running monetr should be able to configure their desired job queue via the config.yaml file.

Job queues that require external scheduling will require some custom legwork for self-hosted environments. In production monetr will be using RabbitMQ for job distribution and Kubernetes CronJobs to trigger scheduled jobs.

---

### High Availability

| Job Engine           | monetr high-availability | high-availability |
|----------------------|--------------------------|-------------------|
| In-Memory            | ❌                        | ❌                 |
| gocraft/work (Redis) | ✅                        | ❌                 |
| RabbitMQ             | ✅                        | ✅                 |

Job engines with a ✅  under `monetr high-availability` can gracefully handle failures of individual monetr processes. This allows for monetr to be updated in a rollout fashion without jobs or job processing being interrupted. Job engines without this will lose any jobs they have in queue, and might lose jobs that are set to be retried or even ones that are in progress. While the chances of data corruption are extremely low as jobs are atomic (changes from a job are only persisted if the entire job succeeds), some jobs might not be performed if monetr loses them.

Job engines with a ✅  under `high-availability` can gracefully handle failures on the monetr (job processing) side, or on the job distribution side. For example, if some instances of RabbitMQ die or are restarted, job processing will continue uninterrupted and no jobs or job data will be lost as long as a quorum is still achievable among the RabbitMQ nodes. But with Redis, if the Redis node that monetr was using for job distribution fails; any jobs not already taken by a monetr instance would be lost. Jobs that were queued up to be retried at a later time would be lost. Inflight jobs should be unaffected as long as the monetr instance the job is running on remains online, but if the monetr instance the job were running on is also interrupted without being able to communicate this failure to Redis, then the job is lost.

---

#### Jobs Converted To New Workflow

- [x] Pull Transactions
- [x] Pull Bank Account Balances
- [x] Process Funding Schedules
- [x] Cleanup Jobs
- [x] Remove Transactions
- [x] Remove Link


pkg/controller/plaid_webhook.go
```diff
@@ -214,13 +215,31 @@ func (c *Controller) processWebhook(ctx iris.Context, hook PlaidWebhook) error {
...
+				LinkId:    link.LinkId,
+				Start:     time.Now().Add(-7 * 24 * time.Hour), // Last 7 days.
+				End:       time.Now(),
+			})
 		case "TRANSACTIONS_REMOVED":
-			_, err = c.job.TriggerRemoveTransactions(link.AccountId, link.LinkId, hook.RemovedTransactions)
+			err = background.TriggerRemoveLink(c.getContext(ctx), c.backgroundJobs, background.RemoveLinkArguments{
+				AccountId: link.AccountId,
+				LinkId:    link.LinkId,
+			})
```
elliotcourant, line 242: This is for transactions_removed, not removing link.

elliotcourant:
This is going to change how jobs are reported to Sentry quite significantly. By the end of this PR all of the sentry code within the background package should be consistent. Even if it does not match what we have right now. 

### Requirements For Sentry

- [ ] Tenant jobs **always** have `accountId` as a tag on the transaction.
- [ ] Panics are reported properly to sentry as failures.
- [ ] Errors also indicate a failed state on the sentry transaction.
- [ ] If possible, find a way to set the transaction that triggered the job transaction.

elliotcourant:
It's unlikely that I will end up implementing the in-memory and RabbitMQ engines as part of this PR as well.

The new structure for background jobs makes it significantly easier to support many different mediums for distributing jobs among the instances. It also isn't feasible for us to run RabbitMQ in acceptance at this time.

If this ends up being the case, I will break up the tasks above into issues for RabbitMQ and in-memory.

elliotcourant:
Due to an error with how the configuration is being loaded, migrations are not being run properly in CI.

```
[test-go] Running go tests for monetr REST API
/usr/local/go/bin/go run github.com/monetr/monetr/pkg/cmd database migrate -d test-db -U api-testing -H postgres
failed to read in config from file: Config File "config" Not Found in "[/root/.monetr /etc/monetr /__w/monetr/monetr]"
2022/01/11 01:34:57 Redis is disabled; cannot use gocraft/work job engine
exit status 1
make: *** [Makefile:210: test-go] Error 1
```

pkg/background/remove_link.go
```diff
@@ -0,0 +1,218 @@
...
+
+	// TODO Add pub/sub stuff
+	//channelName := fmt.Sprintf("link:remove:%d:%d", accountId, linkId)
+	//
+	//if err = r.notify.Notify(span.Context(), channelName, "success"); err != nil {
+	//	log.WithError(err).Warn("failed to send notification about successfully removing link")
+	//	crumbs.Warn(span.Context(), "failed to send notification about successfully removing link", "pubsub", map[string]interface{}{
+	//		"error": err.Error(),
+	//	})
+	//}
```
elliotcourant, line 215: Pubsub for this still needs to be built out.

pkg/background/synchronous.go
```diff
@@ -0,0 +1,77 @@
...
+		marshal: DefaultJobMarshaller,
+	}
+
+	publisher := pubsub.NewPostgresPubSub(log, db)
+
+	jobs := []JobHandler{
+		NewCleanupJobsHandler(log, db),
+		NewProcessFundingScheduleHandler(log, db),
+		NewPullTransactionsHandler(log, db, plaidSecrets, plaidPlatypus, publisher),
+		NewRemoveLinkHandler(log, db),
```
elliotcourant, line 55: The rest of the jobs need to be added here.

Makefile.local
```diff
@@ -81,6 +81,7 @@ mini-info:
 	$(call infoMsg,    Secrets:                                  https://vault.$(LOCAL_DOMAIN))
 	$(call infoMsg,    Redis:                                    https://redis.$(LOCAL_DOMAIN))
 	$(call infoMsg,    gocraft/work UI:                          https://work.$(LOCAL_DOMAIN))
+	$(call infoMsg,    RabbitMQ:                                 https://rabbitmq.$(LOCAL_DOMAIN))
```
elliotcourant, line 84: Since RabbitMQ is not being deployed locally, this should be removed.

pkg/background/cleanup_jobs.go
```diff
@@ -0,0 +1,97 @@
...
+
+	"github.com/getsentry/sentry-go"
+	"github.com/go-pg/pg/v10"
+	"github.com/monetr/monetr/pkg/models"
+	"github.com/pkg/errors"
+	"github.com/sirupsen/logrus"
+)
+
+const (
+	CleanupJobs = "CleanupJobs"
```
elliotcourant, line 15: Right now with the current changes, nothing is ever inserted into the jobs table. With that this job could either be disabled for the time being. Or changes could be made to support the jobs table.

pkg/background/gocraft.go
```diff
@@ -0,0 +1,202 @@
...
+		"queue":     queue,
+		"arguments": arguments,
+	}
+
+	crumbs.Debug(span.Context(), "Enqueueing job for background processing", map[string]interface{}{
+		"queue":     queue,
+		"arguments": arguments,
+	})
+
+	g.log.WithField("queue", queue).Debug("enqueuing job to be run")
```
elliotcourant, line 57: Add a `.WithContext(span.Context())` here.

pkg/background/gocraft.go
```diff
@@ -0,0 +1,202 @@
...
+	select {
+	case <-doneChannel:
+		return nil
+	case <-timeout.C:
+		return errors.New("timeout while draining gocraft/work queue")
+	}
+}
+
+func (g *GoCraftWorkJobProcessor) RegisterJob(ctx context.Context, handler JobHandler) error {
+	log := g.log.WithField("job", handler.QueueName())
```
elliotcourant, line 136: Probably overkill, but would be best to add something here to check if the processor is closed before we do any work.

elliotcourant, line 136: Add a `.WithContext(ctx)` here as well

pkg/background/gocraft.go
```diff
@@ -0,0 +1,202 @@
...
+	log.Trace("registering job handler")
+	g.queue.Job(handler.QueueName(), func(job *work.Job) (err error) {
+		highContext := sentry.SetHubOnContext(context.Background(), sentry.CurrentHub().Clone())
+		span := sentry.StartSpan(highContext, "queue.process", sentry.TransactionName(handler.QueueName()))
+		jobLog := log.WithContext(span.Context())
+		hub := sentry.GetHubFromContext(span.Context())
+		hub.ConfigureScope(func(scope *sentry.Scope) {
+			scope.SetTag("queue", handler.QueueName())
+		})
```
elliotcourant, line 145: Would probably be a good idea to add comments about what all is being done here. This is a bit messy.

pkg/background/gocraft.go
```diff
@@ -0,0 +1,202 @@
...
+	}()
+	select {
+	case <-doneChannel:
+		return nil
+	case <-timeout.C:
+		return errors.New("timeout while draining gocraft/work queue")
+	}
+}
+
+func (g *GoCraftWorkJobProcessor) RegisterJob(ctx context.Context, handler JobHandler) error {
```
elliotcourant, line 135: Add a comment about how to use `RegisterJob`

pkg/background/gocraft.go
```diff
@@ -0,0 +1,202 @@
...
+					hub.ConfigureScope(func(scope *sentry.Scope) {
+						scope.SetLevel(sentry.LevelError)
+					})
+					hub.CaptureException(err)
+				}
+			}
+		}()
+		defer func() {
+			span.Finish()
+		}()
```
elliotcourant, line 172: This was changed to be a func() for debugging, it can be changed back.

pkg/background/gocraft.go
```diff
@@ -0,0 +1,202 @@
...
+					hub.ConfigureScope(func(scope *sentry.Scope) {
+						scope.SetLevel(sentry.LevelFatal)
+					})
+				}
+
+				if err == nil {
+					err = errors.Errorf("panic in job: %v", panicErr)
+				}
+			} else if err != nil {
+				jobLog.Error("error while processing job")
```
elliotcourant, line 161: Leave a comment here about WithError not being used, or add a WithError.

pkg/background/gocraft.go
```diff
@@ -0,0 +1,202 @@
...
+		if scheduledJob, ok := handler.(ScheduledJobHandler); ok {
+			schedule := scheduledJob.DefaultSchedule()
+			log.WithField("schedule", schedule).Trace("job will be run on a schedule automatically")
+			schedulerName := scheduledJob.QueueName() + "GoCraftScheduler"
+			g.queue.Job(schedulerName, func(job *work.Job) error {
+				return scheduledJob.EnqueueTriggeredJob(context.Background(), g.enqueuer)
+			})
+			g.queue.PeriodicallyEnqueue(scheduledJob.DefaultSchedule(), schedulerName)
+		}
+	}
```
elliotcourant, line 199: Add some comments about what is happening here.

pkg/background/job_dist.go
```diff
@@ -0,0 +1,48 @@
...
+
+type TriggerableJobHandler interface {
+	JobHandler
+	// EnqueueTriggeredJob should trigger the actual job using the provided enqueuer. Make sure that the triggered job
+	// can be run on any worker, not necessarily the worker that triggered the job.
+	EnqueueTriggeredJob(ctx context.Context, enqueuer JobEnqueuer) error
+}
+
+type Job interface {
+	Run(ctx context.Context) error
```
elliotcourant, line 37: Add comment here about properly implementing `Run` on a job.

pkg/background/job_dist.go
```diff
@@ -0,0 +1,48 @@
...
+}
+
+type JobProcessor interface {
+	RegisterJob(ctx context.Context, handler JobHandler) error
+	Start() error
+	Close() error
+}
```
elliotcourant, line 44: Add a comment about implementing a job processor

pkg/background/process_funding_schedule.go
```diff
@@ -0,0 +1,277 @@
...
+	}
+
+	log.WithField("count", len(fundingSchedules)).Info("preparing to enqueue funding schedules for processing")
+	crumbs.Debug(ctx, "Preparing to enqueue funding schedules for processing.", map[string]interface{}{
+		"count": len(fundingSchedules),
+	})
+
+	jobErrors := make([]error, 0)
+
+	for _, item := range fundingSchedules {
```
elliotcourant, line 111: Add a comment here about what we are doing.

pkg/background/synchronous.go
```diff
@@ -0,0 +1,77 @@
...
+		marshal: DefaultJobMarshaller,
+	}
+
+	publisher := pubsub.NewPostgresPubSub(log, db)
+
+	jobs := []JobHandler{
+		NewCleanupJobsHandler(log, db),
+		NewProcessFundingScheduleHandler(log, db),
+		NewPullTransactionsHandler(log, db, plaidSecrets, plaidPlatypus, publisher),
+		NewRemoveLinkHandler(log, db, publisher),
```
elliotcourant, line 55: Some job handlers are still missing here

pkg/config/configuration.go
```diff
@@ -295,10 +298,69 @@ type Vault struct {
...
+	// **NOT YET IMPLEMENTED**
+	BackgroundJobEngineRabbitMQ BackgroundJobEngine = "rabbitmq"
+	// BackgroundJobEnginePostgreSQL might be another option for self-hosted deployments. It allows for multiple
+	// replicas of monetr to be deployed and scaled independently. Allowing for a somewhat high-availability deployment.
+	// This job engine will be built out in the future, but has some downsides as it will create a not-insignificant
+	// amount of churn in the PostgreSQL database. Things like vacuuming should be done far more frequently if this job
+	// engine is used.
+	// **NOT YET IMPLEMENTED**
+	BackgroundJobEnginePostgreSQL BackgroundJobEngine = "postgres"
+)
```
elliotcourant, line 340: Create issues for the job engines that are not implemented in this PR.

templates/cronjob.yaml
```diff
@@ -0,0 +1,70 @@
...
+spec:
+  schedule: "0 8 * * *" # TODO Allow this to be changed via values.yaml
+  jobTemplate:
+    metadata:
+      {{- with .Values.podAnnotations }}
+      annotations:
+      {{- toYaml . | nindent 8 }}
+      {{- end }}
+      labels:
+    {{- include "api.selectorLabels" . | nindent 8 }}
```
elliotcourant, line 17: This is not finished, just removee it.

values.yaml
```diff
@@ -101,6 +101,11 @@ databaseMigrations:
 
 customEnvFrom: [ ]
 
+# installCronJobs will generate Kubernetes CronJob resources for triggering job's using an external scheduler. In this
+# case it is basically using Kubernetes as the job scheduler. A single CronJob will be created for each job type that is
+# run on a schedule.
+installCronJobs: false
+
```
elliotcourant, line 108: This isnt finished, just remove it.

PR #1228, opened 2022-11-27
feat!(api): Laying groundwork for multiple funding schedule support.

elliotcourant:
I want to be able to support multiple funding schedules for a variety of
reasons. In order to get there I need to deprecate the existing funding
implementation in regards to spending objects. This migrates the funding
instructions from a one-to-one relationship to a possible one-to-many
relationship.


ui/models/Spending.ts
```diff
@@ -63,7 +61,8 @@ export default class Spending {
   }
 
   getNextContributionAmountString(): string {
-    return formatAmount(this.nextContributionAmount);
+    console.warn("DEPRECATED");
+    return "DEPRECATED";
```
elliotcourant, line 65: Make sure this gets removed

ui/components/Expenses/ExpenseItem.tsx
```diff
@@ -18,9 +21,15 @@ export default function ExpenseItem(props: Props): JSX.Element {
...
   const isSelected = expense.spendingId === selectedExpenseId;
-  const fundingSchedule = useFundingSchedule(expense.fundingScheduleId);
+  const spendingFunding = useSpendingFunding(expense);
+  const fundingSchedules = useFundingSchedules();
+
   const onClick = () => setCurrentExpense(isSelected ? null : expense.spendingId);
 
+  const nextFunding = spendingFunding.result.length > 0 && spendingFunding.result
+    .sort((a, b) => fundingSchedules.get(a.fundingScheduleId).nextOccurrence.unix() > fundingSchedules.get(b.fundingScheduleId).nextOccurrence.unix() ? 1 : -1)[0]
+  const fundingSchedule = nextFunding && fundingSchedules.get(nextFunding.fundingScheduleId);
```
elliotcourant, line 31: Leave a comment about what's going on here.

What happens if we get two funding schedules that land on the same day?

pkg/background/process_funding_schedule.go
```diff
@@ -296,6 +318,13 @@ func (p *ProcessFundingScheduleJob) Run(ctx context.Context) error {
 		return err
 	}
 
+	if len(fundingToUpdate) > 0 {
+		if err = p.repo.UpdateSpendingFunding(span.Context(), p.args.BankAccountId, fundingToUpdate); err != nil {
+			log.WithError(err).Error("failed to update spending funding")
+			return err
+		}
+	}
```
elliotcourant, line 326: No need to wrap with the len, its checked inside the update method.

pkg/background/process_spending.go
```diff
@@ -207,5 +213,15 @@ func (p *ProcessSpendingJob) Run(ctx context.Context) error {
...
-	return errors.Wrap(p.repo.UpdateSpending(span.Context(), p.args.BankAccountId, spendingToUpdate), "failed to update stale spending")
+	if err = p.repo.UpdateSpending(span.Context(), p.args.BankAccountId, spendingToUpdate); err != nil {
+		return errors.Wrap(err, "failed to update stale spending")
+	}
+
+	if len(fundingToUpdate) > 0 {
+		if err = p.repo.UpdateSpendingFunding(span.Context(), p.args.BankAccountId, fundingToUpdate); err != nil {
+			return errors.Wrap(err, "failed to update spending funding")
+		}
+	}
```
elliotcourant, line 224: No need to check len, its checked inside the update method.

pkg/forecast/forecast.go
```diff
@@ -46,11 +46,17 @@ func NewForecaster(spending []models.Spending, funding []models.FundingSchedule)
...
 	for _, spendingItem := range spending {
-		fundingInstructions, ok := forecaster.funding[spendingItem.FundingScheduleId]
-		if !ok {
-			panic("missing funding schedule required by spending object")
+		fundingItems := make([]FundingInstructions, len(spendingItem.SpendingFunding))
+		for i, fundingItem := range spendingItem.SpendingFunding {
+			funding, ok := forecaster.funding[fundingItem.FundingScheduleId]
+			if !ok {
+				panic("missing funding schedule required by spending object")
+			}
```
elliotcourant, line 54: Maybe make this an assert since I added that in this PR too.

pkg/internal/myownsanity/assert.go
```diff
@@ -0,0 +1,11 @@
...
+
+import (
+	"fmt"
+)
+
+func Assert(predicate bool, message string) {
+	if !predicate {
+		panic(fmt.Sprintf("ASSERT FAILED: %s", message))
+	}
+}
```
elliotcourant, line 11: Add a test for this, its stupid easy.

pkg/internal/myownsanity/numbers.go
```diff
@@ -8,6 +8,10 @@ func Int32P(value int32) *int32 {
 	return &value
 }
 
+func Int64P(value int64) *int64 {
+	return &value
+}
+
```
elliotcourant, line 14: Maybe we just make this `NumberP` and use the generic interface below?

pkg/models/spending_funding.go
```diff
@@ -0,0 +1,84 @@
...
+		next := schedule.FundingSchedule.GetNextContributionDateAfter(start, timezone)
+
+		if next.Before(first) || first.IsZero() {
+			first = next
+			funding = schedule
+		}
+	}
+
+	return funding
+}
```
elliotcourant, line 84: This should return an array instead, since we could have two funding schedules land on the same day. Maybe that can be broken out into another PR since this is meant more to just lay the groundwork.

pkg/models/spending_funding.go
```diff
@@ -0,0 +1,84 @@
...
+		dtStart := util.MidnightInLocal(start, timezone)
+		rule.DTStart(dtStart)
+		items := rule.Between(start, end, true)
+		for _, item := range items {
+			contributions[util.MidnightInLocal(item, timezone).Unix()] = struct{}{}
+		}
+	}
+
+	return int64(len(contributions))
+}
```
elliotcourant, line 69: Should add a docstring explaining that this is a unique number of contributions based on date.

If multiple funding schedules were to land on the same day; that is not reflected in the number returned by this function.

pkg/models/spending_funding.go
```diff
@@ -0,0 +1,84 @@
...
+
+	for _, schedule := range s {
+		next := schedule.FundingSchedule.GetNextContributionDateAfter(first, timezone)
+		if next.Before(second) || second.IsZero() {
+			second = next
+		}
+	}
+
+	return first, second
+}
```
elliotcourant, line 50: Write a test where two funding schedules would overlap. This should account for that but I want to be absolutely sure.

pkg/repository/expenses_test.go
```diff
@@ -0,0 +1,66 @@
...
+
+	"github.com/monetr/monetr/pkg/internal/fixtures"
+	"github.com/monetr/monetr/pkg/internal/myownsanity"
+	"github.com/monetr/monetr/pkg/internal/testutils"
+	"github.com/monetr/monetr/pkg/models"
+	"github.com/monetr/monetr/pkg/repository"
+	"github.com/stretchr/testify/assert"
+)
+
+func TestRepositoryBase_GetSpendingById(t *testing.T) {
```
elliotcourant, line 16: Would also be good to add a test that shows we properly get the relation for requesting by funding schedule ID too.

pkg/repository/expenses_test.go
```diff
@@ -0,0 +1,66 @@
...
+			NextContributionAmount: 0,
+		}
+		assert.NoError(t, repo.CreateSpendingFunding(context.Background(), &spendingFunding), "must create spending funding")
+		assert.NotZero(t, spendingFunding.SpendingFundingId, "must have a spending funding ID")
+		assert.NotZero(t, spendingFunding.AccountId, "must have a spending funding ID")
+
+		fullSpending, err := repo.GetSpendingById(context.Background(), bankAccount.BankAccountId, spending.SpendingId)
+		assert.NoError(t, err, "must not return an error reading a valid spending")
+		assert.Equal(t, spending.SpendingId, fullSpending.SpendingId, "should retrieve the same spending")
+		assert.Len(t, fullSpending.SpendingFunding, 1, "should have one spending funding item")
```
elliotcourant, line 63: Should add an assertion that each spending funding object also has a funding schedule associated with it.

pkg/repository/repository.go
```diff
@@ -68,6 +66,16 @@ type BaseRepository interface {
...
+
+	CreateSpendingFunding(ctx context.Context, funding *models.SpendingFunding) error
+	UpdateSpendingFunding(ctx context.Context, bankAccountId uint64, updates []models.SpendingFunding) error
+	GetSpendingFunding(ctx context.Context, bankAccountId uint64, spendingId uint64) ([]models.SpendingFunding, error)
+	ProcessTransactionSpentFrom(
+		ctx context.Context,
+		bankAccountId uint64,
+		input, existing *models.Transaction,
+	) (updatedExpenses []models.Spending, updatedFunding []models.SpendingFunding, _ error)
+	AddExpenseToTransaction(ctx context.Context, transaction *models.Transaction, spending *models.Spending) (*models.SpendingFunding, error)
```
elliotcourant, line 78: Add doc strings.

pkg/repository/transaction.go
```diff
@@ -389,24 +398,45 @@ func (r *repositoryBase) ProcessTransactionSpentFrom(ctx context.Context, bankAc
...
-	return expenseUpdates, r.UpdateSpending(span.Context(), bankAccountId, expenseUpdates)
+	if err = r.UpdateSpending(span.Context(), bankAccountId, expenseUpdates); err != nil {
+		return nil, nil, errors.Wrap(err, "failed to update spending for transaction update")
+	}
+
+	if len(fundingUpdates) > 0 {
+		if err = r.UpdateSpendingFunding(span.Context(), bankAccountId, fundingUpdates); err != nil {
+			return nil, nil, errors.Wrap(err, "failed to update spending funding instructions")
+		}
+	}
```
elliotcourant, line 421: No need for this len.

ui/components/Expenses/ExpenseItem.tsx
```diff
@@ -18,9 +20,15 @@ export default function ExpenseItem(props: Props): JSX.Element {
   } = useStore();
 
   const isSelected = expense.spendingId === selectedExpenseId;
-  const fundingSchedule = useFundingSchedule(expense.fundingScheduleId);
+  const spendingFunding = useSpendingFunding(expense);
+  const fundingSchedules = useFundingSchedules();
+
   const onClick = () => setCurrentExpense(isSelected ? null : expense.spendingId);
 
+  const nextFunding = spendingFunding.result.length > 0 && spendingFunding.result
```
elliotcourant, line 28: We should check if the spendingFunding is loading.

elliotcourant, line 28: What happens if there are multiple "next" fundings on the same day?

ui/components/Expenses/ExpenseItem.tsx
```diff
@@ -58,7 +67,7 @@ export default function ExpenseItem(props: Props): JSX.Element {
           <Typography
             variant="body1"
           >
-            { expense.getNextContributionAmountString() }/{ fundingSchedule?.name }
+            { formatAmount(nextFunding.nextContributionAmount) }/{ fundingSchedule?.name }
```
elliotcourant, line 70: This shows `NaN` when we are still loading stuff in.

ui/hooks/spending.ts
```diff
@@ -106,6 +107,9 @@ export function useUpdateSpending(): (_spending: Spending) => Promise<void> {
             previous.map(item => item.spendingId === updatedSpending.spendingId ? updatedSpending : item),
         ),
         queryClient.invalidateQueries(`/bank_accounts/${ updatedSpending.bankAccountId }/balances`),
+        queryClient.invalidateQueries(
+          `/bank_accounts/${ updatedSpending.bankAccountId }/spending/${ updatedSpending.spendingId }/funding`,
+        ),
```
elliotcourant, line 112: We could do both of these in a single invalidate.

ui/hooks/spending.ts
```diff
@@ -106,6 +107,9 @@ export function useUpdateSpending(): (_spending: Spending) => Promise<void> {
...
         queryClient.invalidateQueries([`/bank_accounts/${ updatedSpending.bankAccountId }/forecast/next_funding`]),
```
elliotcourant, line 113: The query key for this one is an array, but can we just use the string? If we can we can make these 3 invalidates a single one.

pkg/background/process_funding_schedule.go
```diff
@@ -254,18 +255,36 @@ func (p *ProcessFundingScheduleJob) Run(ctx context.Context) error {
 					continue
 				}
 
+				spendingFunding, err := p.repo.GetSpendingFunding(span.Context(), spending.BankAccountId, spending.SpendingId)
```
elliotcourant, line 258: This might result in some doubled up queries with `GetSpendingByFundingSchedule` where we hit the spending funding table twice?

ui/hooks/spending.ts
elliotcourant: Need to test this before merging, but its updated.

pkg/controller/funding_schedules.go
```diff
@@ -232,10 +239,31 @@ func (c *Controller) putFundingSchedules(ctx iris.Context) {
...
 					return
 				}
+
+				if updatedFunding != nil {
+					if err = repo.UpdateSpendingFunding(c.getContext(ctx), bankAccountId, []models.SpendingFunding{
+						*updatedFunding,
+					}); err != nil {
+						c.wrapPgError(ctx, err, "failed to update spending funding")
+						return
+					}
```
elliotcourant, line 265: There are a ton of these, maybe I should just make a helper function that takes a single spending funding object that is nullable?

And if nil is passed to it, then the function is a no-op; but if something is passed to it then it does the update?

pkg/models/spending.go
```diff
@@ -163,19 +161,19 @@ func (e *Spending) CalculateNextContribution(
...
 	if e.NextRecurrence.Before(now) {
 		e.LastRecurrence = &e.NextRecurrence
 		e.NextRecurrence = nextRecurrence
 	}
 
-	return nil
+	// Update the funding item with our calculated contribution amount.
+	nextFunding := funding.GetNextFunding(now, timezone)
+	nextFunding.NextContributionAmount = totalContributionAmount
+	return &nextFunding, nil
```
elliotcourant, line 178: This wont work with multiple funding schedules that overlap. Maybe I should tinker with ripping out this entire bit of code and we just use the forecasting package? It would create a circular import but I could work around that.

But that way the calculations are exactly the same for forecasting and for the actual calculation. The only difference being the forecast assumes something is always spent on the day specified. Where as the real use case isn't concerned with spending (except when it comes to funding).

elliotcourant: Updated this so now it does work with multiple. Not using the forecasting code yet, it still needs work and should not be relied on just yet.

pkg/forecast/spending.go
```diff
@@ -200,37 +193,58 @@ func (s *spendingInstructionBase) getNextSpendingEventAfter(ctx context.Context,
...
+		case nextRecurrence.Before(fundingFirst.Date):
+			// No-op, goals do nothing when they are completed.
+			return nil
+		case nextRecurrence.Equal(fundingFirst.Date):
+			// Next event will the completion of the goal.
+			event.Date = nextRecurrence
+			event.ContributionAmount = totalContributionAmount
+			event.RollingAllocation = event.RollingAllocation + totalContributionAmount
+			event.Funding = fundingFirst.Events
+			event.CompletedGoal = event.RollingAllocation >= s.spending.TargetAmount
```
elliotcourant, line 247: Just a thought, should goals and expenses be calculated the same way at all?

Should we have separate spending instructions for each type?

pkg/repository/transaction.go
```diff
@@ -389,24 +396,43 @@ func (r *repositoryBase) ProcessTransactionSpentFrom(ctx context.Context, bankAc
...
+
+	if err = r.UpdateSpending(span.Context(), bankAccountId, expenseUpdates); err != nil {
+		return nil, nil, errors.Wrap(err, "failed to update spending for transaction update")
+	}
+
+	if len(fundingUpdates) > 0 {
+		if err = r.UpdateSpendingFunding(span.Context(), bankAccountId, fundingUpdates); err != nil {
+			return nil, nil, errors.Wrap(err, "failed to update spending funding instructions")
+		}
```
elliotcourant, line 416: No need for length check

ui/components/Expenses/ExpenseItem.tsx
```diff
@@ -18,9 +20,12 @@ export default function ExpenseItem(props: Props): JSX.Element {
   } = useStore();
 
   const isSelected = expense.spendingId === selectedExpenseId;
-  const fundingSchedule = useFundingSchedule(expense.fundingScheduleId);
+  const spendingFunding = useSpendingFunding(expense);
+
   const onClick = () => setCurrentExpense(isSelected ? null : expense.spendingId);
 
+  const { funding, schedule } = spendingFunding.next.length > 0 && spendingFunding.next[0] || { };
```
elliotcourant, line 27: Should really just solve for multiple funding schedules in the UI now.

This will easily break in the future otherwise.

If there were two on the same day, then the funding.nextContributionAmount would be 1/2 what it is supposed to be (as shown by the UI)

PR #1284, opened 2023-01-06
deps(ui): material-ui upgrade + fixes.

elliotcourant:
- deps(ui): update material-ui monorepo
- fix(ui): Fixing material UI upgrade issues.


elliotcourant:
Fixes https://github.com/monetr/monetr/pull/1012

PR #1305, opened 2023-01-08
feat(dev): Building out new build system.

elliotcourant:


elliotcourant:
Ultimately monetr will move to CMake for builds in the future.

PR #1370, opened 2023-03-16
feat(ui): Improving appearance of authentication screens.

elliotcourant:


elliotcourant:
image

instead of

image

PR #1393, opened 2023-05-15
feat(ui): Use rspack instead of webpack going forward.

elliotcourant:


.storybook/main.ts
```diff
@@ -24,44 +17,10 @@ const config: StorybookConfig = {
...
-    name: '@storybook/react-webpack5',
-    options: {
-      fastRefresh: true,
-    },
+    name: 'storybook-react-rspack',
   },
   docs: {
     autodocs: 'tag',
   },
-  webpackFinal: async config => {
```
elliotcourant, line 35: @th3fallen do you think that its worth trying to tackle rspack with storybook right now?

Otherwise if we just fix jest then this is basically ready to go.

And we'd just need to use webpack for a little while for storybook only.

package.json
```diff
@@ -2,7 +2,59 @@
...
+  "scripts": {
+    "build": "rspack build",
+    "build:staging": "MONETR_ENV=staging rspack build --mode production",
+    "build:production": "yarn build --mode production",
+    "build-dev": "rspack --mode development",
+    "build-prod": "rspack --mode production",
+    "clean": "rm -rf build/* || true",
+    "start": "MONETR_ENV=${MONETR_ENV:=local} rspack serve --mode development",
+    "test": "jest",
+    "coverage": "jest --coverage --collectCoverageFrom=ui/**/*.{js,jsx,ts,tsx} --coverageDirectory=./reports",
```
elliotcourant, line 14: Since the collect coverage from is specified in the package.json for jest is the flag even necessary?

package.json
```diff
@@ -2,7 +2,59 @@
...
   "version": "0.1.0",
   "private": true,
-  "type": "module",
+  "scripts": {
+    "build": "rspack build",
+    "build:staging": "MONETR_ENV=staging rspack build --mode production",
+    "build:production": "yarn build --mode production",
+    "build-dev": "rspack --mode development",
+    "build-prod": "rspack --mode production",
+    "clean": "rm -rf build/* || true",
```
elliotcourant, line 11: Should just proxy to `make clean` instead

package.json
```diff
@@ -116,10 +123,10 @@
     "@storybook/addon-viewport": "7.0.11",
     "@storybook/blocks": "7.0.11",
     "@storybook/react": "7.0.11",
-    "@storybook/react-webpack5": "7.0.11",
     "@storybook/test-runner": "0.10.0",
     "@storybook/testing-library": "0.0.14-next.2",
     "@swc/core": "1.3.40",
+    "@swc/jest": "0.2.26",
```
elliotcourant, line 129: Whats this do? @th3fallen

elliotcourant (commented):
@th3fallen I left some more comments, these arent all for you but some are questions that youll know more about than I.

elliotcourant:
Also we can remove `.swcrc` from this

PR #1478, opened 2023-07-15
feat(build): Adding support for bazel.

elliotcourant:
Right now monetr's build process is really complicated, especially for release builds. It is possible to build monetr
using only go build, but the resulting binary would not have the third party notices, embedded icon packs or embedded
user interface.

Right now some complex make tasks suplement this, but making these work on everyone's machine is difficult. Making them
consistent is even harder, and the requirements to build a release worthy binary of monetr from scratch is difficult.

I'm hoping to improve this with bazel. Bazel will only replace the effort of building the release binary. All other
things like local development will still be done with make or other tooling.


elliotcourant:
no idea what to do; works locally just fine, but fails to build because its missing files somehow in CI.

thought the entire point of bazel was the build was consistent everywhere 😬

elliotcourant:
Build is actually working now, but tests are not. Probably because its not receiving the postgres hostname from the environment variable?

elliotcourant:
Making some progress, the build is now working:
- UI correctly embeds into the binary
- Icons are now also embedding properly
- Version and revision information embedding properly.

---

Running into an issue now though where the binary is not resolving the postgres DB host properly?

elliotcourant:
Also squashed commits, was up to 33; this will make rebasing onto main easier in the future as i tinker more.

PR #1489, opened 2023-08-31
feat(docs): Improving documentation development workflow.

elliotcourant:
This is going to be breaking out the documentation container workflow
such that it can be run separately from the rest of the local
development workflows.


elliotcourant:
@JayRKyd When you have some time if you want to give this branch a look. I've updated mkdocs to the latest version as well as updated the documentation around actually _running_ the doc site on windows. Give it a try and let me know if you run into any issues.

JayRKyd:
okay i’ll try it out tomorrow and give you some feedback

PR #1549, opened 2023-10-06
feat(jobs): Adding support for the neoq job processor.

elliotcourant:
This adds support for the neoq job processor backed by PostgreSQL. There
are still things that need to be tweaked in neoq before this can be
fully merged. But this will be a good starting point for that work.


elliotcourant:
Don't delete this branch, I've removed neoq from the primary code base because of a licensing issue.

PR #1551, opened 2023-10-12
feat(ui): Tinkering with react data router.

elliotcourant:


elliotcourant:
The whole point of this was to do scroll restoration; but i did it myself https://github.com/monetr/monetr/commit/24b2161ff11fab742922944f09900ba5298d7a4a

PR #1562, opened 2023-10-16
feat(ci): Significantly improve CI/CD caching.

elliotcourant:


elliotcourant:
Significantly is a stretch, something weird is happening with the golang build cache where it is taking a _really_ long time to build the test binaries.

elliotcourant:
the first commit on this PR took 15 minutes before I killed it and barely made any progress.

but this commit re-run a second time was already nearly half way through tests in just 2m 30s, something is strange with the caching here.

elliotcourant:
Note: I did also test throwing out the precompiled binary entirely and just running go test directly. But this was an order of magnitude slower even on a 16 core machine.

PR #1576, opened 2023-10-17
feat(ui): Migrating email templates to react-email

### LINKED ISSUE #1568: feat(email): Generate email templates using react email
Right now the email templates are basically static and were generated by copy and pasting some HTML code out of a wysiwyg editor. The HTML for them is awful and really hard to maintain if they ever need to be changed.

Email templates should be generated via https://react.email/ instead. Then the HTML itself is only compiled at build time instead of statically version controlled.

elliotcourant:
Email templates will be generated using react-email going forward. This
will make it easier to maintain emails as well as style and customize
them as time goes on. As well as keep email appearance consistent
between different message types.

---

### TODO

- [x] Verify email address template
- [x] Forgot password template
- [ ] Password changed?
- [x] Update template structs in Go.
- [x] CMake task to build HTML from React.
- [x] Tailwind wrapper component, need to load monetr's tailwind config instead of whatever react-email is using by default.
- [ ] Test? Idk what tests would even look like?
- [ ] Building a preview of emails would be neat as part of CI/CD.

---

### Follow ups for later

- [ ] Send emails asynchronously, right now they are sent synchronously which can be bad for HTTP requests and does not give us an easy way to track individual email failures outside of basic logging. It would be nice to kick emails into an async job queue; maybe it doesn't need to be distributed via the background package but that way a slow to send email does not block the HTTP request. And a failure can be handled outside of the scope of the HTTP call.
- [ ] PGP signature? A lot of email clients don't even support this.

---

Resolves #1568


elliotcourant:
react email creates a directory named `.react-email` in whatever directory it is run in. 

I'd rather not have another directory to clean up? If possible I should open a PR to tweak this:

https://github.com/resendlabs/react-email/blob/main/packages/react-email/source/utils/constants.ts

to make it so that the directory it uses can be passed as a flag. Or make this cmake task run relative to the binary directory and just provide an absolute path to the email dir. This way the temp dir it creates does not pollute the working directory.

elliotcourant:
> react email creates a directory named `.react-email` in whatever directory it is run in.
> 
> I'd rather not have another directory to clean up? If possible I should open a PR to tweak this:
> 
> https://github.com/resendlabs/react-email/blob/main/packages/react-email/source/utils/constants.ts
> 
> to make it so that the directory it uses can be passed as a flag. Or make this cmake task run relative to the binary directory and just provide an absolute path to the email dir. This way the temp dir it creates does not pollute the working directory.


It looks like that the directory is only created when youre running the preview thing locally. So I don't think its really that important. I can make a cmake target that puts it out of tree for now or move it entirely into a container, but not a huge deal.

elliotcourant:
Will need to address this: https://app.snyk.io/org/elliotcourant/project/2dddc993-5296-4673-b98e-6817bd5a7f6b/pr-check/3f37fed7-4839-4e10-af12-96670b72d26b

emails/VerifyEmailAddress.tsx
```diff
@@ -0,0 +1,101 @@
...
+  lastName?: string;
+  emailAddress?: string;
+  supportEmail?: string;
+  inviteLink?: string;
+}
+
+export const VerifyEmailAddress = ({
+  baseUrl = 'https://my.monetr.dev', // '{{ .BaseURL }}',
+  firstName = '{{ .Login.FirstName }}',
+  lastName = '{{ .Login.LastName }}',
```
elliotcourant, line 30: These both need to be urlescaped

elliotcourant: and html escaped

PR #1607, opened 2023-11-08
fix(dev): Fixing local development getting stuck.

elliotcourant:


elliotcourant:
This does not solve the actual problem; that will be addressed here: https://github.com/monetr/monetr/issues/1608

But it does fix the CI/CD

PR #1628, opened 2023-11-20
feat(api): Laying groundwork for detecting recurring transactions.

elliotcourant:
This PR lays the groundwork for detecting recurring transactions. It has the individual components (I think) I'll need in order to do so:
- Group similar transactions by their name, allow fuzzy matching, synonyms throwing out useless words.
- Detect similarities in transactions by their name individually using text comparison algorithms.
- Define the recurrence windows to search for.

More work still needs to be done to actually put these pieces together and leverage them. But this is a solid foundation for the detecting.


server/recurring/cluster.go
```diff
@@ -0,0 +1,271 @@
...
+			vector[index] = tfidfValue
+		}
+		var norm float64
+		for _, value := range vector {
+			norm += value * value
+		}
+		norm = math.Sqrt(norm)
+		for i, value := range vector {
+			vector[i] = value / norm
+		}
```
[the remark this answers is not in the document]
elliotcourant: Also if I don't need a ton of precision I might be able to switch to float32 for all of these as well. That would let the SIMD code process 8 items at a time instead of 4.

server/recurring/cluster.go
```diff
@@ -0,0 +1,271 @@
...
+	Transaction Transaction
+	Vector      []float64
+}
+
+func (a Datum) Distance(b Datum) float64 {
+	var distance float64
+	for i, value := range a.Vector {
+		distance += math.Pow(value-b.Vector[i], 2)
+	}
+	return distance
```
elliotcourant, line 173: Could also be done with SIMD

```asm
TEXT ·sumOfSquaredDifferences(SB), NOSPLIT, $0
    MOVQ a+0(FP), SI            // Load the address of a.Vector
    MOVQ b+24(FP), DX           // Load the address of b.Vector
    MOVQ a+8(FP), CX            // Load the length of the vectors
    VXORPD YMM0, YMM0, YMM0     // Clear the accumulator register YMM0

    // Loop over the vectors in batches of 4 (since we're using YMM registers)
    LOOP:
        VMOVUPD (SI), YMM1      // Load 4 float64 elements from a.Vector
        VMOVUPD (DX), YMM2      // Load 4 float64 elements from b.Vector
        VSUBPD YMM2, YMM1, YMM1 // Compute the difference (a-b)
        VMULPD YMM1, YMM1, YMM1 // Square the differences
        VADDPD YMM1, YMM0, YMM0 // Accumulate the squared differences

        ADDQ $32, SI            // Move to the next 4 elements in a.Vector
        ADDQ $32, DX            // Move to the next 4 elements in b.Vector
        SUBQ $4, CX             // Decrement the loop counter (4 elements processed)
        JNZ LOOP                // Continue loop if there are more elements

    // Finalize the sum of squared differences
    // Extract the high 128 bits and add them to the low 128 bits
    VEXTRACTF128 $0x1, YMM0, XMM1
    VADDSD XMM1, XMM0, XMM0
    VHADDPD XMM0, XMM0, XMM0

    // Store the result
    VMOVSD XMM0, ret+48(FP)

    RET
```

elliotcourant: A visualization of the registers:

```
 YMM0, YMM1, YMM2 register (256 bits total):
+------------------+------------------+------------------+------------------+
|  float64 value 0 |  float64 value 1 |  float64 value 2 |  float64 value 3 |
+------------------+------------------+------------------+------------------+
```

server/recurring/cluster.go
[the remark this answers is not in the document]
elliotcourant: instructions are backwards

server/recurring/cluster.go
elliotcourant: instructions are backwards

server/recurring/cluster.go
```diff
@@ -0,0 +1,305 @@
...
+
+import (
+	"math"
+	"sort"
+	"strings"
+)
+
+var (
+	specialWeights = map[string]float64{
+		"amazon": 10,
```
elliotcourant, line 11: Add a `pwp` multiplier for `0`. This will make it so the word is essentially useless.

Also add something where if a words value is 0 we maybe exclude it from the vector entirely?

server/internal/calc/euclidean_amd64.s
```diff
@@ -0,0 +1,29 @@
...
+  MOVQ a+8(FP), DX  // Load the length of a into the DX register.
+  MOVQ a+0(FP), AX  // Load the pointer of the first array.
+  MOVQ b+24(FP), BX // Load the pointer of the second array.
+
+  VXORPD Y0, Y0, Y0 // Clear the accumulator registers.
+
+  LOOP:
+    VMOVUPD (AX), Y1  // Load the current 4 float64s from a into the 256-bit register Y1.
+    VMOVUPD (BX), Y2  // Load the current 4 float64s from b into the 256-bit register Y2.
+    VSUBPD Y2, Y1, Y1 // Subtract the 4 from a from the 4 from b and store the result in Y1.
```
elliotcourant, line 14: Is this actually subtracting a from b? I think the first two registers are reversed.

elliotcourant:
Going to need this: 

https://github.com/klauspost/cpuid

monetr for amd64 will be compiled with the SIMD functions, but if the actual host running the code (for whatever reason) does not have AVX, then it needs to fallback to using the pure go version.

server/recurring/cluster.go
```diff
@@ -0,0 +1,335 @@
...
+		// Then start constructing a cluster around this point.
+		newCluster := d.expandCluster(index, neighbors)
+		// Set the cluster's unique ID to the lowest numeric ID in that cluster.
+		// HACK: I need a way to uniquely identify each cluster. Generally by using the contents of that cluster. This
+		// relies on the contents of that cluster remaining consistent over time. While the order of the clusters might
+		// change in the future or they might expand as new transactions show up, I need to know which cluster they get
+		// added to in order to tune things over time. This has the potential to cause issues on its own, what if the
+		// cluster algorithm changes enough that the "lowest ID" gets kicked out of the cluster? What if we push a bad
+		// change and the clusters change entirely? Or what if that "lowest ID" gets moved to a different cluster. This
+		// needs improvement, but I think this should be fine for the initial implementation of the clustering algorithm.
```
elliotcourant, line 256: I think the best way would instead be to identify each cluster by the highest value word in that cluster.

PR #1641, opened 2023-12-15
feat: Improve recurring transaction detection.

elliotcourant:
This code is an experiment on isolating specific amounts within a known set of similar transactions.

The goal of this is to be able to cherry pick specific transactions out of a cluttered dataset to identify them as recurring.

An example of this is amazon, there might be tons of amazon transactions but only a handful of them are actually something like an "amazon prime subscription". So the goal with this is to isolate those transactions that are part of a subscription.

### DIFF
--- a/server/recurring/amount_test.go
+++ b/server/recurring/amount_test.go
@@ -285,18 +285,37 @@ func TestAmountDifferent(t *testing.T) {
 			8419:  1,
 			9018:  1,
 		}
+		// freq = map[int64]int64{
+		// 	1500: 10,
+		// 	1700: 15,
+		// }
 		fmt.Sprint(freq)
-		data = make([]float64, 0)
-		for amount, count := range freq {
-			for i := 0; i < int(count); i++ {
-				data = append(data, float64(amount))
-			}
-		}
+		// data = make([]float64, 0)
+		// for amount, count := range freq {
+		// 	for i := 0; i < int(count); i++ {
+		// 		data = append(data, float64(amount))
+		// 	}
+		// }
 
 		sort.Float64s(data)
 
 		// Estimate the bandwidth
 		bandwidth := SilvermansRuleOfThumb(data)
+
+		bandwidths := make([]float64, 0)
+		for i := 500; i < 5000; i += 10 {
+			bandwidths = append(bandwidths, float64(i))
+		}
+		scores := make([]float64, len(bandwidths))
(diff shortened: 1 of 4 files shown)


server/recurring/amount_test.go
```diff
@@ -0,0 +1,397 @@
...
+
+		sort.Float64s(data)
+
+		// Estimate the bandwidth
+		bandwidth := SilvermansRuleOfThumb(data)
+
+		bandwidths := make([]float64, 0)
+		for i := 500; i < 5000; i += 10 {
+			bandwidths = append(bandwidths, float64(i))
+		}
```
elliotcourant, line 308: Should also append the bandwidth from silvermans rule of thumb here too and then sort the array.

We generate several bandwidths starting at $5.00 and then going up by 10 cents, but including the silvermans rule might cause this to be even more accurate since it has no such constraint. OR we might be able to tune our own increment to be 50 cents or even $1 based instead.

elliotcourant:
# TODO

Need to implement some kind of data smoothing. I'm thinking a Gaussian smoothing might be the best for the dataset since its also reasonable to implement in go:

```go
func gaussianKernel(size int, sigma float64) []float64 {
	kernel := make([]float64, size)
	sum := 0.0
	m := size / 2

	for i := 0; i < size; i++ {
		diff := float64(i - m)
		kernel[i] = math.Exp(-(diff * diff) / (2 * sigma * sigma))
		sum += kernel[i]
	}

	// Normalize the kernel
	for i := range kernel {
		kernel[i] /= sum
	}

	return kernel
}

func gaussianSmooth(data []float64, sigma float64) []float64 {
	size := int(sigma * 6) // a common choice for kernel size
	if size%2 == 0 {
		size++ // ensure kernel size is odd
	}

	kernel := gaussianKernel(size, sigma)
	halfSize := size / 2
	smoothedData := make([]float64, len(data))

	for i := range data {
		var weightedSum float64
		var weightSum float64

		for j := -halfSize; j <= halfSize; j++ {
			if i+j >= 0 && i+j < len(data) {
				weight := kernel[halfSize+j]
				weightedSum += data[i+j] * weight
				weightSum += weight
			}
		}

		smoothedData[i] = weightedSum / weightSum
	}

	return smoothedData
}
```

This should also improve peak detection.

Need to experiment with various sigma values, or should the sigma be determined by the bandwidth value?

PR #1642, opened 2023-12-16
feat(api): Implement PostgreSQL backed job queue.

elliotcourant:
This provides a new job queue implementation as an alternative to
redis/gocraft. This job queue is designed to be more resistant to
failures like processing a job more than once.


elliotcourant:
Cron jobs are not consumed yet. TODO

PR #1655, opened 2024-01-03
feat(api): Redo MFA to use scoped tokens.

elliotcourant:
This is an effort to make MFA dependent on the audience of the token
that was issued rather than a weird state where credentials must be
resubmitted.

### DIFF
--- a/server/controller/authentication.go
+++ b/server/controller/authentication.go
@@ -155,7 +155,7 @@ func (c *Controller) postLogin(ctx echo.Context) error {
 		crumbs.IncludeUserInScope(c.getContext(ctx), user.AccountId)
 
 		// Check if the login requires MFA in order to authenticate.
-		if login.TOTP != "" {
+		if login.TOTPEnabledAt != nil {
 			log.Debug("login requires TOTP MFA")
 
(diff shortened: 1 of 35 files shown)


server/controller/authentication.go
```diff
@@ -234,6 +243,60 @@ func (c *Controller) loginEndpoint(ctx echo.Context) error {
...
+	repo := c.mustGetAuthenticatedRepository(ctx)
+	me, err := repo.GetMe(c.getContext(ctx))
+	if err != nil {
+		c.updateAuthenticationCookie(ctx, ClearAuthentication)
+		return c.wrapAndReturnError(ctx, err, http.StatusInternalServerError, "unable to retrieve current user")
+	}
+
+	if err := me.Login.VerifyTOTP(request.TOTP, c.clock.Now()); err != nil {
+		return c.returnError(ctx, http.StatusUnauthorized, "invalid TOTP code")
+	}
```
elliotcourant, line 263: Double check to see if this will remove the cookie, we need to keep that around. We dont want to remove it if they get it wrong, it should be good for a few minutes.

PR #1658, opened 2024-01-07
feat(plaid): Separate Plaid data in the database.

elliotcourant:
Right now every record has some Plaid specific fields tied to it that
makes it difficult to manage or add other types of data sources since
Plaid is essentially embeded in the data model. This is an effort to
break Plaid into its own data tree and then tie those Plaid records back
to the records individually. This will allow us to adopt a pattern where
each data source has its own tree of data that then ties back to a
single record. It will also allow us to potentially have two data
sources for the same thing. For example manual + Plaid, or Privacy +
manual, etc.

It also opens the door for us to more easily add support for other data
sources like Stripe Financial Connections or Teller.


server/repository/link.go
```diff
@@ -52,7 +51,9 @@ func (r *repositoryBase) GetNumberOfPlaidLinks(ctx context.Context) (int, error)
 
 	count, err := r.txn.ModelContext(span.Context(), &models.Link{}).
 		Where(`"link"."account_id" = ?`, r.accountId).
-		Where(`"link"."link_type" = ?`, models.PlaidLinkType).
+		// TODO This is partially correct, but we should really look at the status
+		// on the plaid link itself for this?
+		Where(`"link"."plaid_link_id" IS NOT NULL`).
```
elliotcourant, line 56: This should be partiall reverted

server/repository/link.go
```diff
@@ -71,7 +72,7 @@ func (r *repositoryBase) GetLinkIsManual(ctx context.Context, linkId uint64) (bo
 	ok, err := r.txn.ModelContext(span.Context(), &models.Link{}).
 		Where(`"link"."account_id" = ?`, r.AccountId()).
 		Where(`"link"."link_id" = ?`, linkId).
-		Where(`"link"."link_type" = ?`, models.ManualLinkType).
+		Where(`"link"."plaid_link_id" IS NULL`).
```
elliotcourant, line 75: This should be reverted

server/repository/link.go
```diff
@@ -95,7 +96,7 @@ func (r *repositoryBase) GetLinkIsManualByBankAccountId(ctx context.Context, ban
 		JoinOn(`"bank_account"."link_id" = "link"."link_id" AND "bank_account"."account_id" = "link"."account_id"`).
 		Where(`"link"."account_id" = ?`, r.AccountId()).
 		Where(`"bank_account"."bank_account_id" = ?`, bankAccountId).
-		Where(`"link"."link_type" = ?`, models.ManualLinkType).
+		Where(`"link"."plaid_link_id" IS NULL`).
```
elliotcourant, line 99: revert

PR #1688, opened 2024-02-23
feat(docs): Embed the user interface in the docs directly

elliotcourant:


elliotcourant:
This actually seems to work now, but getting data into the dummy components is difficult. I think itd be nice to have a way to capture live data somehow and use that?

elliotcourant:
https://storybook.js.org/docs/writing-stories/mocking-data-and-modules/mocking-modules#setting-up-and-cleaning-up

elliotcourant:
- Update stories to use mock axios instead of mock service worker
- Update story fixture data
- Re-add the story component back into the user interface

PR #1716, opened 2024-03-29
feat(ui): Migrating UI tests to use bun instead.

elliotcourant:


interface/src/pages/login.spec.tsx
```diff
@@ -1,29 +1,46 @@
...
+// //   useNavigate: () => mockUseNavigate,
+// // }));
+//
+// mock.module('react-router-dom', () => {
+//   return {
+//     // __esModule: true,
+//     // ...require('react-router-dom'),
+//     useNavigate: () => mockUseNavigate,
+//   };
+// });
```
elliotcourant, line 24: Still figuring out how to mock a hook from a module

elliotcourant:
Need to start working on this again, bun tests are so much faster; and moving away from mock service worker will also solve some other issues. Like being able to re-embed the UI in the docs site.

elliotcourant:
Need to add code coverage from bun still 

https://bun.sh/docs/test/coverage

PR #1736, opened 2024-04-09
feat(api): Building out a sourcemap system for sentry

elliotcourant:
I want to be able to bundle the golang source code in the sentry SDK at
build time. This is progress towards that.


elliotcourant:
Need to build out the following implementations:

https://github.com/getsentry/sentry-go/blob/master/sourcereader.go

https://github.com/getsentry/sentry-go/blob/9311a060b6e1d6f2c0bb783789bebcc88b873764/integrations.go#L215-L328

PR #1740, opened 2024-04-12
feat(api): Migrating to ULID instead of 64-bit integer IDs

elliotcourant:
This will make writing data in bulk easier in the future, it will also
make this data more portable between environments or between databases.


elliotcourant:
I think instead of generating the initial ULIDs in go they should be done via this: https://github.com/geckoboard/pgulid

This way the migration can be done as a single SQL migration rather than multiple.

elliotcourant:
The code side is complete, now its just fixing tests or fixing code depending on how the test fails.

Also the fixtures I had generated a long time ago for forecasting need to be regenerated somehow.

PR #1773, opened 2024-05-28
feat(api): Adding QFX processing to import transactions

elliotcourant:
Actually adding the processor to import transactions from QFX files.


server/background/process_qfx_upload.go
```diff
@@ -0,0 +1,467 @@
...
+	})
+	channel := fmt.Sprintf("account:%s:transaction_upload:%s:progress", args.AccountId, args.TransactionUploadId)
+	payload := string(status)
+	if err := h.publisher.Notify(ctx, channel, payload); err != nil {
+		return errors.Wrap(err, "failed to send progress notification for job")
+	}
+	log.WithFields(logrus.Fields{
+		"channel": channel,
+		"payload": payload,
+	}).Trace("sent progress notification for file upload")
```
elliotcourant, line 106: Should move the notification to send after the row is updated. Otherwise there might be a race condition in the client where they receive a progress notification and read the upload record, only to get stale data.

PR #1833, opened 2024-07-07
Fix/1601: Scroll restoration

yeinerfonseca:
This adds a reusable hook to restore positioning in any route with an scrollable element.
The scroll positioning will be restored regardless of the navigation type.

An enhancement could be implemented for the following scenario:
Say we have route `foo/bar`, if it accessed through a POP navigation type, then scroll is restored
to the the last positioning. Otherwise, if navigation type is PUSH, then scroll is restored to zero (top).

Let me know your thoughts.

### DIFF
--- a/interface/src/hooks/useScrollRestoration.tsx
+++ b/interface/src/hooks/useScrollRestoration.tsx
@@ -1,0 +1,99 @@
+import React from 'react';
+import { useLocation } from 'react-router-dom';
+
+interface ScrollPositionProviderProps {
+  children: React.ReactNode;
+}
+
+interface IScrollPositionContext {
+  saveScrollPosition: (key: string, position: number) => void;
+  getScrollPosition: (key: string) => number;
+}
+
+const ScrollPositionContext = React.createContext<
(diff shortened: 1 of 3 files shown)


interface/src/hooks/useScrollRestoration.tsx
```diff
@@ -0,0 +1,99 @@
...
+  const context = React.useContext(ScrollPositionContext);
+  if (!context) {
+    throw new Error(
+      'useScrollPosition must be used within a ScrollPositionProvider',
+    );
+  }
+  return context;
+}
+
+export function ScrollPositionProvider({
```
elliotcourant, line 27: This entire thing is fantastic, I need to play around with it for a bit. But I really love this implementation.

elliotcourant (commented):
I really like how you implemented the scroll restoration in such a way that it can easily be reused by basically any component. Thank you!

I'll play around with this locally here soon, but i want to get this merged since it'll simplify a lot of stuff!

I'll take a deeper look into the code then, and merge/comment from there!

PR #1839, opened 2024-07-14
feat(dev): Local development improvements

elliotcourant:
This is just me messing around with improving the local development
environment. It adds a feature flag service that can be used in
development.

It also has some basic changes towards:

- Better support for non-tls environments.
- Actually giving each service a subdomain.


elliotcourant:
https://docs.flipt.io/configuration/telemetry#configuration-file

Disable telemetry for local dev

PR #1842, opened 2024-07-17
Fix unstable QueryClient issue:#1841

### LINKED ISSUE #1841: bug(ui): React query client not stable
https://github.com/monetr/monetr/blob/ce6ba51bc57bb5631fbf5f817b72f27dc037f96d/interface/src/components/MQueryClient.tsx#L11-L31

https://tanstack.com/query/latest/docs/eslint/stable-query-client

The query client should be declared as a useState or should be declared outside the component so that re-rendering does not recreate the query client.

Er-Sadiq:
Title: Fix: Stabilize QueryClient by Memoizing its Creation

Description:
This PR addresses the issue where the QueryClient was being recreated on every render of the MQueryClient component. By using the useMemo hook, the QueryClient instance is now memoized, ensuring it is only created once and persists across renders. This improves performance and stability, resolving the instability issues previously observed.

Changes:

Memoized the creation of the QueryClient instance in MQueryClient.tsx using the useMemo hook.
Issue Fixed:

Fixes #1841
Testing:

Verified that the QueryClient is not recreated on each render.
Ensured the application behaves correctly and performance is improved.
Notes:

No breaking changes introduced.
Added comments to explain the memoization logic.
Please review the changes and let me know if there are any questions or further adjustments needed. Thank you!

### DIFF
--- a/interface/src/components/MQueryClient.tsx
+++ b/interface/src/components/MQueryClient.tsx
@@ -1,4 +1,4 @@
-import React from 'react';
+import React, { useMemo } from 'react';
 import { QueryClient, QueryClientProvider } from '@tanstack/react-query';
 
 import { ReactElement } from './types';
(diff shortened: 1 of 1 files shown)


elliotcourant:
Does useMemo do that, or would useState be better like in their docs? I'm not actually familiar with the nuances and differences between the two for a component like this which receives no props of its own, and therefore shouldn't frequently (if ever?) re-render?

Overall this looks good though, I'll give it a try today and make sure things work!

elliotcourant (approved):

PR #1933, opened 2024-09-05
feat(billing): Reconcile subscriptions periodically

### LINKED ISSUE #1463: bug(stripe): Better handling of not receiving webhook.
Plaid handle's not receiving webhooks pretty well at this point, by periodically performing a sync without having been pinged. This way it is not reliant on webhooks for self hosting.

Stripe really needs to be the same way, and while that wouldn't have prevented the issue with webhooks outright failing. It would have made it so customers could still access their application in the mean time.

---

Stripe subscriptions should be checked via a cron every so often, this way we aren't relying only on a webhook.

Maybe we also increase the grace period from 24 hours to something like 72. But check the subscription status at 48 hours. We typically receive webhooks within 24 hours of the subscription cycle ending, this would give us a healthy buffer and allow webhooks to outright fail and still maintain subscription statuses. We could also alert that we are updating subscriptions via a cron when that happens so we can check to see if webhooks are broken.

elliotcourant:
This adds a background job that runs every 12 hours and looks for
accounts who should have received a webhook for their subscription
within those past 12 hours. If it finds any, it will then kick off a
follow up job to reconcile that accounts subscription against Stripe's
API. This way even if monetr misses a webhook from Stripe, we can still
maintain an accurate billing state in monetr, and we don't accidently
lock user's out of their accounts.

Resolves #1463


elliotcourant:
- [ ] Test setting up a new subscription again
- [ ] Test reconciling a subscription.
- [ ] Add command to force recon.

PR #2051, opened 2024-10-25
feat(build): Moving storybook to webpack and interface to rsbuild

elliotcourant:
I don't want to use rsbuild for storybook and rspack for the actual UI.
I'd rather have consistent tooling; and webpack and rspack are very
consistent between each other.

This moves everything back to plain old webpack 5, which should
hopefully also make it easier to upgrade to a newer version of
storybook.


stories/main.ts
```diff
@@ -38,11 +45,75 @@ const config: StorybookConfig = {
...
+        useSWC: true, 
+      }, 
+      jsc: {
+        transform: {
+          react: {
+            development: isDevelopment,
+            refresh: isDevelopment,
+          },
+        },
+      },
```
elliotcourant, line 60: This is definitely not propagating to SWC.

elliotcourant:
As an alternative to this; https://rsbuild.dev/guide/basic/configure-rspack I could just actually go down the rsbuild rabbit hole and just cram rspack in there lol

elliotcourant:
Because holy _fuck_ webpack is slow

elliotcourant:
So this isn't entirely complete yet, I would like to move storybook over to using rsbuild as well; but as part of that migration i think it just makes sense to move it to the latest version of storybook as well.

So assuming the build passes, im just going to merge this as is. It removes rspack from the storybook and moves interface over to rsbuild.

PR #2075, opened 2024-10-27
feat(math): Adding fourier caluclations and notes

elliotcourant:
I think that I could build a recurring transaction detection thing that
uses a fourier transform to find out if and how frequently a group of
transactions recurr.

This is an initial commit which contains a shitload of notes as I try to
understand the algorithm (which might not be implemented correctly). My
end goal with this is to implement a pure go version as well as an X86
assembly version(s) leveraging both AVX and AVX512 features to perform
the math.


elliotcourant:
https://godbolt.org/z/a8j43s4dv Current playground

elliotcourant:
https://github.com/gonum/gonum/blob/master/dsp/fourier/fourier.go < Example of a similar implementation;

server/internal/calc/fourier.go
```diff
@@ -0,0 +1,96 @@
...
+	}
+
+	fftEven := FastFourierTransform(even)
+	fftOdd := FastFourierTransform(odd)
+
+	result := make([]complex128, n)
+	for k := 0; k < n/2; k++ {
+		// -2 * math.Pi comes out to:
+		// MOVSD   $f64.c01921fb54442d18(SB), X1 = -6.283185307179586
+		t := complexExponential(-2*math.Pi*float64(k)/float64(n)) * fftOdd[k]
```
elliotcourant, line 61: So the parameters here make it so that this could actually be recalculated up to a point and then stored and simply used.

elliotcourant: 1. Precalculate all of these values up to a certain point `N`.
2. Group the values into various sizes, like 128-bit, 256-bit and 512-bit. Ideally this way the precalculated values can simply be moved into a XMM, YMM or ZMM register directly as a single instruction. Then the current register for `fttOdd` can be multiplied by that other register.

So if the complex exponential was simple like `1, 2, 3, 4, 5, 6, 7, 8`. Then for AVX512 we'd have a fttOdd register like

`11, 12, 13, 14, 15, 16, 17, 18` or 8x 64 bit floats. each 128 bit pair is a complex 128.

Then we can do
```
VMULPD Z1, Z2, Z3
```

Where Z1 is the fft odd register; and Z2 is our complex exponential.

So the complex exponential register would look like:

```
1, 1, 2, 2, 3, 3, 4, 4
```

Because the values in the Z1 register are each 128 bits. So we need to multiply each 128 bit pair by the same complex exponential.

elliotcourant: I wonder if I could use `VMOVUPD` to just cram the precalculated complex exponentials into a 512 bit register. I would need to have them already layed out in pairs though like i have above with 1,1 2,2 3,3 etc.

server/recurring/recurring.go
```diff
@@ -1,190 +1,174 @@
...
+			scores[f] = score
+			continue
+		}
 
-	if len(amountsSorted) <= 16 {
-		return amountsSorted
+		value := result[int(index)]
+		real := real(value)
+		imaginary := imag(value)
+		magnitude := math.Sqrt((real * real) + (imaginary * imaginary))
```
elliotcourant, line 157: Might want to use `cmpx.Abs` instead? Not sure what the difference is.

elliotcourant:
Probably going to address the assembly parts in a different PR

elliotcourant:
TLDR; this works really well. The PR includes tests that are sufficient to cover all the cases that I care about at the moment. But I'm not going to dive into the assembly optimizations of the math just yet. I also have not reimplemented the recurring transactions controller atm. I'll add those in follow up PRs.

This also includes some improvements to the TFIDF code that should improve transaction clustering.

PR #2109, opened 2024-11-05
feat: Allow transactions to be created via dialog

elliotcourant:
This is to add a modal that would allow users to create transactions
manually without a file import.


elliotcourant:
Need to allow transactions to be deleted too

PR #2301, opened 2025-01-07
feat(billing): Allow users to subscribe before their trial ends

elliotcourant:
If someone wants to subscribe to monetr before their trial has ended
they will now be able to do so. This will not trample the trial data,
but will make the subscription status take priority. It also makes sure
that we don't send trial expiry notifications when we don't need to.


elliotcourant:
I've tested this locally but I'm not going to merge yet because I want to think through it some more. I've missed things with billing in the past and have broken stuff so I don't want to rush this.

PR #2358, opened 2025-01-25
feat(ui): Allow currency of bank account to be changed

### LINKED ISSUE #2330: Allow an account's currency to be changed but warn users that this may be a destructive action. When moving between currencies that are similarly represented (like CAD to USD) there should be no issues, but when moving between currencies which are not at all similarly represented (JPY and USD for example) this could corrupt all of the data stored in monetr for that account.

elliotcourant:
This is the last piece needed for someone to actually customize the
currency of a manual bank account in monetr.

This will also include some quality of life changes around the bank
account picker, including moving the "Add account" button into the
actual bank account list which seems to make more sense.

Resolves #2330

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -159,7 +159,7 @@ require (
 	golang.org/x/exp v0.0.0-20240222234643-814bf88cf225 // indirect
 	golang.org/x/oauth2 v0.24.0 // indirect
 	golang.org/x/sync v0.10.0 // indirect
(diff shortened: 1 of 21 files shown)


elliotcourant:
https://pkg.go.dev/golang.org/x/text@v0.21.0/language#Confidence

elliotcourant:
https://pkg.go.dev/golang.org/x/text@v0.21.0/language#Tag

PR #2452, opened 2025-02-17
feat(api): Adding PATCH functionality to API

elliotcourant:
This change is an effort to support more complex API calls with ease via
patch requests to the API. Golang is not terribly good at this so it
requires a substantial amount of boiler plate. But this change should
make it easy to create more PATCH endpoints going forward as well as
abstract a lot of the logic away.


elliotcourant:
https://github.com/darccio/mergo/issues/138

```
accountId=acct_01jmaq1ypc53psvgvfxz0aqngs error=Failed to patch request data: type mismatch on AvailableBalance field: found string, expected int64 file=/home/runner/work/monetr/monetr/server/controller/controller.go:126 func=github.com/monetr/monetr/server/controller.(*Controller).reportError level=error loginId=lgn_01jmaq1yp96efpcjd8ztfaecx5 requestId=ef8d7c51-5134-4c23-96b5-1f5eccdd2690 test=TestPatchBankAccount/happy_path_patch_a_manual_link_bank_account userId=user_01jmaq1ypevag1qmtczw8j62fq msg=error in request: Failed to patch request data: type mismatch on AvailableBalance field: found string, expected int64
```

> error in request: Failed to patch request data: type mismatch on AvailableBalance field: found string, expected int64

Because the field is a `json.Number` it cannot map to a field that is a strict number

This is due to the code in: https://github.com/darccio/mergo/blob/2b1eb9c67d7332f286430af241180c5005a6a5a4/map.go#L114

elliotcourant:
I could create a custom transformer that was a bit more friendly based on the scan code from go-pg: https://github.com/go-pg/pg/blob/bce00adc25696cc87008ccc4daf10f6350bbdae8/types/scan_value.go#L89

This way things like string to string or similar base types would be friendly

PR #2692, opened 2025-09-17
fix(job): Fixing trial notification expiry

### LINKED ISSUE #2682: bug(trial): Trial expiry notification job is duplicating work
We aren't sending out emails when we aren't supposed to, but we do keep trying to check to send an email notification for logins that don't need them.

```loki
{app="monetr"} | json | queue=`NotificationTrialExpiry` | loginId=`lgn_...`
```

For a single login ID that has not verified its email address; we are getting the same expiry job every day.

The query should be updated to consider verified emails.

elliotcourant:
Resolves #2682


elliotcourant:
# TODO

- [ ] Add another test case that makes sure that we do not enqueue a job for accounts that were created without a verified email address from the owner.

Issue #2427, opened 2025-02-15
feat(plaid): Support European and UK institutions


elliotcourant:
This issue supersedes https://github.com/monetr/monetr/issues/2378 as it will track specifically work towards supporting EU and UK instituions.

# Current Status

At this point in time monetr _does not_ support EU and UK institutions in the hosted version. It is also almost impossible to support them in the self hosted version as well only because the price to access these institutions requires monthly minimums with Plaid that are prohibitively expensive.

# Alternatives

I am currently working on alternate ways to support EU and UK institutions without needing to go through Plaid. Whether this means exploring other providers such as [Tink](https://demo.tink.com/), [TrueLayer](https://truelayer.com/) or via https://github.com/monetr/monetr/issues/2333. As well as looking into supporting common file formats provided by EU and UK institutions in https://github.com/monetr/monetr/issues/2352.

# Help out

**If you are in the EU or UK and would like to help add support please reach out! ([Contact Us](https://monetr.app/contact/)).**

I am currently looking for the following:

### Information on common file formats

Obviously CSV and XLSX formats are extremely common, but I am still scoping work for those formats as every bank provides slightly different data in their files making it very difficult to support.

https://github.com/monetr/monetr/issues/2352 is at the top of my list at the moment for file formats to support based on feedback from a user in the Netherlands.

If there are other file formats that your institution supports please let me know!

### Direct Connections

What institutions support direct connections? This option might not be as viable for the hosted version of monetr as I'm still navigating the regulatory space of the EU when it comes to processing financial data of end users. But for self-hosted options there should be little issue as long as the person has credentials.

### Something else?

If I'm missing something or if you have other ideas for how to approach this **please reach out!** More than anything I just need information.

GregSutcliffe:
I would be interested in helping here, at least for self-hosted. I currently run [FireFlyIII](https://github.com/firefly-iii/firefly-iii) which uses GoCardless for it's bank integrations. My bank is HSBC (UK, if it matters) which has the most *basic* format for transactions (just date,description,value per account).

elliotcourant:
I'll investigate GoCardless a bit as well, it looks like it supports a few US institutions; so hopefully I can test it out here and then support it for EU/UK institutions no problem. Thank you @GregSutcliffe !

elliotcourant:
[...]

PR #2857, opened 2025-12-11
feat(api): Adding Lunch Flow as a datasource

elliotcourant:
Lunch Flow provides a way for users who are self hosting their own
budgeting apps to sync their data with institutions that are in the
EU/UK as well as other countries as well. This was previously very
difficult to do as an individual.

This is adding support for Lunch Flow as a datasource for monetr, but it
will only be available for self hosted instances.

Related to #2427

### DIFF
--- a/server/repository/link.go
+++ b/server/repository/link.go
@@ -66,30 +66,6 @@ func (r *repositoryBase) GetNumberOfPlaidLinks(ctx context.Context) (int, error)
 	return count, nil
 }
 
-func (r *repositoryBase) GetLinkIsManual(
-	ctx context.Context,
-	linkId ID[Link],
-) (bool, error) {
-	span := crumbs.StartFnTrace(ctx)
-	defer span.Finish()
-	span.SetData("linkId", linkId)
-
-	ok, err := r.txn.ModelContext(span.Context(), &Link{}).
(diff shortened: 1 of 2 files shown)


teynar:
Hey @elliotcourant! Thanks for working on adding support to Lunch Flow, it was me who got in contact with Amr since I saw that you wanted support for EU and UK institutions and I was excited to try Monetr since Plaid is not available in my region.

I was checking your commits and I saw that you were using `https://lunchflow.com` as the base URL, but I think that the new base url is `https://lunchflow.app`, I think that @amrawadk forgot to update the url in the API docs, because I tried doing an API call and the lunchflow.com domain does not have an api (it is listed as for sale), could you please check Amr? Thanks to you both!

elliotcourant:
Okay the lunch flow sync is actually working now, I'm still working on polishing off the changes and the language used in some of the setup UI.

But it is now actually working :tada: I'm hoping to have this merged and released soon!

amrawadk:
> Okay the lunch flow sync is actually working now, I'm still working on polishing off the changes and the language used in some of the setup UI.
> 
> But it is now actually working 🎉 I'm hoping to have this merged and released soon!

Exciting indeed! Let me know if we could be of any help 😄

PR #3101, opened 2026-04-12
refactor(email): Email templates without react-email

elliotcourant:
This is a migration away from react-email, instead leveraging rsbuild to compile email templates from scratch. This
gives me more control over how the templates are built as well as continues monetr's migration away from tailwindcss.

This also trims monetr's dependency tree even further.


interface/src/hooks/useMountEffect.ts
```diff
@@ -1,5 +1,5 @@
 import { type EffectCallback, useEffect } from 'react';
 
 export default function useMountEffect(callback: EffectCallback) {
-  useEffect(callback, []);
+  useEffect(callback, [callback]);
```
elliotcourant: Should test or revert this

emails/package.json
```diff
@@ -2,10 +2,23 @@
...
+    "@rsbuild/core": "2.0.0-rc.1",
+    "@rsbuild/plugin-react": "2.0.0-rc.2",
+    "@rsbuild/plugin-sass": "1.5.1",
+    "@types/html-to-text": "9.0.4",
+    "@types/react": "18.3.26",
+    "@types/react-dom": "18.3.7",
+    "cheerio": "1.2.0",
+    "html-to-text": "9.0.5",
+    "juice": "11.0.1",
     "react": "18.3.1",
```
elliotcourant, line 20: With this, interface/ is now the last holdout for react 19

emails/src/preview/templates.ts
```diff
@@ -0,0 +1,14 @@
+import type { EmailTemplate } from '@monetr/emails/templates';
+import { templates } from '@monetr/emails/templates';
+
+export interface TemplateEntry {
+  name: string;
+  component: EmailTemplate;
+  previewProps: Record<string, any>;
```
elliotcourant: Can be `unknown`

emails/src/rspack.d.ts
```diff
@@ -0,0 +1,15 @@
+// Ambient types for rspack/webpack require.context
+interface RequireContext {
+  keys(): string[];
+  <T = any>(id: string): T;
```
elliotcourant: can also be unknown

emails/src/toPlainText.ts
```diff
@@ -0,0 +1,16 @@
+import { convert as htmlToText } from 'html-to-text';
+
+export function toPlainText(html: string): string {
+  return htmlToText(html, {
+    selectors: [
+      { selector: 'img', format: 'skip' },
+      { selector: '[data-skip-in-text]', format: 'skip' },
```
elliotcourant, line 7: Could strip the data-skip-in-text attribute too since we already strip class

elliotcourant: Added this

PR #3124, opened 2026-04-18
feat(api): Schema validation

elliotcourant:
This is part of the work to get API keys working, if I want to support
proper API clients then I need proper schema validation on EVERYTHING
that is client facing.


elliotcourant:
Will require:

- https://github.com/Oudwins/zog/pull/219
- https://github.com/Oudwins/zog/pull/218

to be merged before this can be finalized.

Nil fields breaks spendingId -> free to use

omit empty is just a footgun, not critical but would be nice to not trip over it

PR #3127, opened 2026-04-20
feat: Automated expense/funding transactions

### LINKED ISSUE #3125: feat: Auto create transactions for manual budgets
For manual links, allow the option to automatically create transactions for:

- Expenses, on the day the expense is due create a transaction for that amount that affects balance and "spend" that transaction from it's associated expense.
- Funding schedules, same as expense; but create a deposit transaction for the estimated amount of the funding schedule.

Both of these affect balance, and are good for expenses that are fixed amounts or funding that is fixed.

---

This might introduce friction with varying expenses/funding where the amount of the deposit or transaction is modified AFTER it is automatically created.

elliotcourant:
If you are using a manual budget, expenses and funding schedules can now
automatically create transactions when they recur. This way if you are
manually entering things, you don't need to manually enter these
transactions every time.

This can be enabled per expense and per funding schedule, and is only
supported on manual links.

Resolves https://github.com/monetr/monetr/issues/3125

### DIFF
--- a/interface/src/modals/NewFundingModal.tsx
+++ b/interface/src/modals/NewFundingModal.tsx
@@ -1,7 +1,7 @@
 import { Fragment, useCallback, useId, useRef } from 'react';
 import NiceModal, { useModal } from '@ebay/nice-modal-react';
 import { startOfDay, startOfTomorrow } from 'date-fns';
-import type { FormikHelpers } from 'formik';
+import { type FormikHelpers, useFormikContext } from 'formik';
 import { useSnackbar } from 'notistack';
 
 import type { ApiError } from '@monetr/interface/api/client';
@@ -35,7 +35,6 @@ interface NewFundingValues {
 
 function NewFundingModal(): JSX.Element {
   const switchId = useId();
-  const autoCreateSwitchId = useId();
   const { inTimezone } = useTimezone();
   const modal = useModal();
   const ref = useRef<MModalRef>(null);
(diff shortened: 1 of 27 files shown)


interface/src/modals/NewFundingModal.tsx
```diff
@@ -142,6 +152,30 @@ function NewFundingModal(): JSX.Element {
...
+                    >
+                      Auto create transaction
+                    </label>
+                    <p className='text-sm text-dark-monetr-content'>
+                      Automatically add a deposit transaction for the estimated deposit each time the funding schedule
+                      would occur.
+                    </p>
+                  </div>
+                  <Switch
+                    checked={values.autoCreateTransaction}
```
elliotcourant: This should always show on manual links, but should be disabled when estimated deposit is 0.

server/funding/funding_jobs/process_funding_schedules.go
```diff
@@ -164,9 +178,54 @@ func ProcessFundingSchedule(ctx queue.Context, args ProcessFundingScheduleArgume
...
+
+				txn := models.Transaction{
+					BankAccountId: args.BankAccountId,
+					// Credits are stored as negative values in monetr.
+					Amount:       -*fundingSchedule.EstimatedDeposit,
+					Date:         fundingDate,
+					Name:         fundingSchedule.Name,
+					OriginalName: fundingSchedule.Name,
+					IsPending:    false,
+					Source:       models.TransactionSourceManual,
```
elliotcourant: Should it be indicated somewhere that these transactions were created automatically?

PR #3152, opened 2026-04-29
refactor(api): Adding transaction import groundwork

elliotcourant:
Transaction imports will succeed transaction uploads, where imports will
ultimately support more file types in the end.

---

This PR adds more groundwork for transaction imports, specifically to support CSV file imports as a first pass.

It trims some of the features in the mapping spec (reducing row number by X things for complexity reasons).

It also adds the table struct in order to take a mapping spec and a CSV file input and turn that into meaningful data.

There is still a lot of work to be done here, like how does preview data ultimately get stored and served back to the UI? How long is preview data considered valid?

### DIFF
--- a/server/controller/transaction_imports_test.go
+++ b/server/controller/transaction_imports_test.go
@@ -60,6 +60,7 @@ func TestPostTransactionImport(t *testing.T) {
 		response.Status(http.StatusOK)
 		response.JSON().Path("$.bankAccountId").String().IsEqual(bank.BankAccountId.String())
 		response.JSON().Path("$.fileId").String().NotEmpty()
+		response.JSON().Path("$.delimeter").IsEqual(",")
 		response.JSON().Path("$.headers").Array().IsEqual([]string{"Date", "Description", "Amount", "Balance"})
 	})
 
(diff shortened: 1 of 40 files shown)


server/controller/transaction_imports_test.go
```diff
@@ -0,0 +1,533 @@
...
+			WithPath("bankAccountId", bank.BankAccountId).
+			WithMultipart().
+			WithFileBytes("data", "transactions.csv", csvData).
+			WithCookie(TestCookieName, token).
+			Expect()
+
+		response.Status(http.StatusOK)
+		response.JSON().Path("$.bankAccountId").String().IsEqual(bank.BankAccountId.String())
+		response.JSON().Path("$.fileId").String().NotEmpty()
+		response.JSON().Path("$.headers").Array().IsEqual([]string{"Date", "Description", "Amount", "Balance"})
```
elliotcourant, line 64: Add another assertion here to make sure that we derive the correct delimeter.

server/controller/transaction_imports_test.go
```diff
@@ -0,0 +1,533 @@
...
+				Expect()
+
+			response.Status(http.StatusNotFound)
+			response.JSON().Path("$.error").String().IsEqual("failed to retrieve transaction import: record does not exist")
+		}
+	})
+}
+
+func TestPatchTransactionImport(t *testing.T) {
+	t.Run("selecting a mapping kicks off the preview job", func(t *testing.T) {
```
elliotcourant, line 384: Add test suite for:
- Going from an invalid status to another invalid status
- Patching an import that doesn't belong to the current user
- Patching a field that isn't allowed or supported

server/datasources/csv/csv_jobs/preview_csv_import.go
```diff
@@ -0,0 +1,148 @@
...
+		// Load the bank account ahead of processing the file, we need this for
+		// currency data and will use it for balance updates later.
+		j.bankAccount, err = j.repo.GetBankAccount(ctx, args.BankAccountId)
+		if err != nil {
+			return errors.Wrap(err, "failed to retrieve bank account for file import sync")
+		}
+
+		// Load the file and its data into memory.
+		if err := j.loadFile(ctx); err != nil {
+			return err
```
elliotcourant, line 149: Would be nice to log the rows here until we are storing them properly.

server/datasources/csv/header.go
```diff
@@ -0,0 +1,56 @@
...
+	}
+
+	for _, delimeter := range testDelimeters {
+		csvReader := csv.NewReader(bytes.NewReader(preview))
+		csvReader.Comma = delimeter
+		csvReader.TrimLeadingSpace = true
+
+		header, err := csvReader.Read()
+		if err != nil {
+			// TODO Log the error?
```
elliotcourant, line 42: If we pass a logger here, it would be good to accumulate the errors and then if none of the readers succeed we can log all the errors to help server admins debug their uploads.

server/datasources/csv/header.go
```diff
@@ -0,0 +1,56 @@
...
+			continue
+		}
+
+		// We must have at least:
+		// - Date
+		// - Amount
+		// - Description
+		if len(header) >= 3 {
+			return delimeter, header, buffer, nil
```
elliotcourant, line 51: Ditto here for logs

server/datasources/table/table.go
```diff
@@ -0,0 +1,292 @@
...
+type Row struct {
+	RowNumber int       `json:"rowNumber"`
+	ID        string    `json:"id"`
+	Amount    int64     `json:"amount"`
+	Memo      string    `json:"memo"`
+	Merchant  *string   `json:"merchant,omitempty"`
+	Date      time.Time `json:"date"`
+	Posted    bool      `json:"posted"`
+	Balance   int64     `json:"balance"`
+}
```
elliotcourant, line 28: Should add a validate function here too to make sure that the resulting data is not goofy. Like validating that the memo is printable; same with ID and merchant.

PR #3198, opened 2026-05-10
refactor(api): Deprecate ReCAPTCHA

elliotcourant:
This deprecation removes all support for ReCAPTCHA, if you have it
configured on your server then this should not break anything. ReCAPTCHA
will simply not function after this update. ReCAPTCHA is not being
replaced by anything, it is recommended that operators instead
instrument rate limiting via nginx or whatever reverse proxy they are
using, or leverage cloudflare or something else to mitigate spam if they
feel the need to.


elliotcourant:
I will be merging this before the v1.15.0 release.

PR #3234, opened 2026-05-20
refactor(api): Self-managed database migrations

elliotcourant:
This is part of a larger effort to eventually migrate entirely from
go-pg to uptrace/bun. Bun has its own migration management but switching
from go-pg to bun invisibly is tricky. By rolling my own migration
system it will make the transition to bun easier since the table
tracking the migrations and the format of the migrations will not need
to change. This will also make it easier to eventually add support for a
separate track of SQLite migrations.


server/migrations/functional/2021050999_TestGoMigration.go
elliotcourant, line 1: This was always a bogus/test migration. Maybe someday monetr will support pure go migrations but I doubt it.

server/migrations/doc.go
```diff
@@ -0,0 +1,26 @@
...
+//  2. Use .tx.up.sql when the body can run inside a transaction. That's
+//     what we want for basically everything. Reach for plain .up.sql only
+//     when Postgres won't let you wrap the statement in a transaction:
+//     CREATE INDEX CONCURRENTLY, some CREATE EXTENSION variants,
+//     ALTER TYPE ... ADD VALUE, and so on.
+//  3. Optionally add a matching .tx.down.sql or .down.sql with the inverse
+//     statements. We validate down files for shape at startup but never
+//     actually run them, they're just there as documentation.
+//  4. Rebuild. //go:embed picks the new file up automatically, no code
+//     changes required.
```
elliotcourant, line 20: At some point I'll make a CMake script that can be run like `make new-migration` that will automatically generate this file and name it properly.

server/migrations/migrations.go
```diff
@@ -2,87 +2,116 @@ package migrations
...
+	for _, f := range files {
+		if f.Version > latest {
+			latest = f.Version
+		}
 	}
 
-	if _, _, err := collection.Run(db, "init"); err != nil {
-		return nil, errors.Wrap(err, "failed to initialize schema migrations")
+	exec := newPgExecutor(db)
+	if err := ensureSchemaTable(context.Background(), log, exec); err != nil {
```
elliotcourant: Might be better to just pass ctx through here.

server/migrations/migrations.go
```diff
@@ -2,87 +2,116 @@ package migrations
...
+		db:            db,
+		exec:          exec,
+		files:         files,
+		latestVersion: latest,
 	}, nil
 }
 
+// CurrentVersion returns the highest version recorded in schema_migrations,
+// or 0 if no migrations have been applied yet.
 func (m *MonetrMigrationsManager) CurrentVersion() (int64, error) {
```
elliotcourant: Ditto here

server/migrations/runner.go
```diff
@@ -0,0 +1,260 @@
...
+// support:
+//
+//	YYYYMMDDNN_Name.up.sql
+//	YYYYMMDDNN_Name.tx.up.sql
+//	YYYYMMDDNN_Name.down.sql
+//	YYYYMMDDNN_Name.tx.down.sql
+//
+// Names may contain underscores (2021041102_Balances_View is real) but not
+// dots, since we use dots to mark the tx/direction suffix.
+var migrationFilenameRegex = regexp.MustCompile(`^(\d+)_([^.]+)\.(tx\.)?(up|down)\.sql$`)
```
elliotcourant: Should make sure the date prefix is also a fixed size, its tripped things up before.

Also a bad migration should panic such that we catch this at build/test time.

server/migrations/runner.go
```diff
@@ -0,0 +1,260 @@
...
+	Filename      string
+}
+
+func parseMigrationFilename(name string) (migrationFile, error) {
+	match := migrationFilenameRegex.FindStringSubmatch(name)
+	if match == nil {
+		return migrationFile{}, errors.Errorf("unrecognized migration filename %q", name)
+	}
+
+	version, err := strconv.ParseInt(match[1], 10, 64)
```
elliotcourant, line 56: We should really check the length of the match here.

server/migrations/runner.go
```diff
@@ -0,0 +1,260 @@
...
+	files []migrationFile,
+) (oldVersion, newVersion int64, err error) {
+	if err := exec.Exec(ctx, "SELECT pg_advisory_lock(?)", advisoryLockKey); err != nil {
+		return 0, 0, errors.Wrap(err, "failed to acquire migration advisory lock")
+	}
+	defer func() {
+		if unlockErr := exec.Exec(ctx, "SELECT pg_advisory_unlock(?)", advisoryLockKey); unlockErr != nil {
+			log.WarnContext(ctx, "failed to release migration advisory lock", "err", unlockErr)
+		}
+	}()
```
elliotcourant, line 148: Should I also capture an advisory lock for the create table too?

server/migrations/runner.go
```diff
@@ -0,0 +1,260 @@
...
+
+// gopgExistsSQL is true if the legacy gopg_migrations table is in the current
+// schema. We scope to current_schema() so a gopg_migrations sitting in some
+// other schema (e.g. a shared public in a multi-tenant cluster) doesn't
+// trigger a seed copy into the wrong place.
+const gopgExistsSQL = `SELECT EXISTS (
+    SELECT 1 FROM information_schema.tables
+    WHERE table_schema = current_schema()
+      AND table_name = 'gopg_migrations'
+)`
```
elliotcourant, line 245: Build this using the query builder instead.

server/migrations/runner.go
```diff
@@ -0,0 +1,260 @@
...
+//     TODO Remove this filter (and the rest of the gopg seed path) once we're
+//     confident no production DB still has gopg_migrations.
+//   - ON CONFLICT makes it idempotent across restarts and tolerant of
+//     partially-seeded state.
+const seedFromGopgSQL = `INSERT INTO schema_migrations (version, applied_at)
+SELECT version, COALESCE(created_at, NOW())
+FROM gopg_migrations
+WHERE version > 0
+  AND version <> 2021050999
+ON CONFLICT (version) DO NOTHING`
```
elliotcourant, line 262: Ditto query builder

elliotcourant: Also select for update?

server/migrations/runner_test.go
```diff
@@ -0,0 +1,160 @@
...
+	require.Len(t, ups, 1)
+}
+
+func TestDiscoverMigrations_RealEmbed(t *testing.T) {
+	ups, err := discoverMigrations(embeddedMigrations)
+	require.NoError(t, err)
+
+	// 99 .tx.up.sql + 6 .up.sql = 105 up files in the production embed.
+	// If a new migration is added, bump this assertion.
+	require.Len(t, ups, 105)
```
elliotcourant: This means the test needs to be updated every time a new migration is added. While this is a nice insight to have to prevent a regression; this isn't terribly helpful going forward. Would be better to say "Greater than"

server/migrations/schema_test.go
```diff
@@ -1,13 +1,316 @@
...
+// testutils itself imports server/migrations and would form a cycle. The
+// log helpers do come from testutils/testlog, which sits in its own sub-
+// package precisely to keep itself cycle-free.
+
+func testPgOptions(t *testing.T) *pg.Options {
+	return &pg.Options{
+		Network: "tcp",
+		Addr: net.JoinHostPort(
+			myownsanity.CoalesceStrings(os.Getenv("MONETR_PG_ADDRESS"), os.Getenv("POSTGRES_HOST"), "localhost"),
+			myownsanity.CoalesceStrings(os.Getenv("MONETR_PG_PORT"), os.Getenv("POSTGRES_PORT"), "5432"),
```
elliotcourant: If we're doing this anyway we should also do:

https://www.postgresql.org/docs/current/libpq-envars.html

And make it so that we also inherit the env vars that psql itself would use.

server/migrations/schema_test.go
```diff
@@ -1,13 +1,316 @@
...
+// nothing in it yet, no migrations applied. Cleanup drops the database when
+// the test ends.
+func newCleanDatabase(t *testing.T) *pg.DB {
+	parentOpts := testPgOptions(t)
+	parent := pg.Connect(parentOpts)
+	require.NoError(t, parent.Ping(context.Background()))
+
+	// Postgres truncates identifiers at 63 bytes. 30 hex chars (15 bytes of
+	// the hash) is comfortably unique per test name and well under the cap.
+	hash := fmt.Sprintf("%x", sha256.Sum256([]byte(t.Name())))
```
elliotcourant: Might be better to use crc then or fnv instead?

PR #3268, opened 2026-05-30
refactor(api): Improving transaction validation

elliotcourant:
This is a proof of concept PR for doing schema validation using just
monetr/validation instead of zog or json schemas.

It also surfaces errors returned from the API for schema validation back
into the form that submitted them, allowing for improved error feedback
to users in some situations.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -27,7 +27,7 @@ require (
 	github.com/jarcoal/httpmock v1.4.1
 	github.com/labstack/echo/v4 v4.15.2
 	github.com/monetr/devslog v0.0.17
-	github.com/monetr/validation v1.0.6
+	github.com/monetr/validation v1.1.0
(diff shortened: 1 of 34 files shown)


elliotcourant:
Okay I'll have some changes to push here shortly but this is how its looking so far:


- Schemas are going to end up being stateless validation, is the field's value with no context considered valid?
- Stateless also means a lack of awareness of time which will require removing some min(time.now) things.
- Validation that happens after the request is parsed that **IS** stateful will not return a full array of errors. It will instead return a simple `error: "message"` format. However some errors on some endpoints may return additional fields/data as necessary.

This way schemas can be made for every endpoint reasonably, and things like validating a schema is easy. But validating data against something meaningful like data in the database or a config value is still manual.

PR #3272, opened 2026-05-31
refactor(ui): Migrating to typescript v6

elliotcourant:


interface/src/components/funding/FundingItem.tsx
```diff
@@ -27,6 +27,10 @@ export default function FundingItem(props: FundingItemProps): JSX.Element {
   const rule = rrulestr(funding.ruleset);
   const letter = funding.name.toUpperCase().charAt(0) || '?';
 
+  if (!locale || !localeCurrency) {
+    return null;
+  }
```
elliotcourant, line 32: If I can do something here where this will always return a default locale or the locale specific rendering is offloaded to its own component then that will make this less dumb

interface/src/components/funding/FundingItem.tsx
elliotcourant, line 82: Like it would be kind of dumb, but making this its own component and then doing a skeleton here would be way better.

interface/src/components/funding/FundingTimeline.tsx
```diff
@@ -78,10 +79,14 @@ export default function FundingTimeline(props: FundingTimelineProps): JSX.Elemen
   );
 }
 
-function TimelineItem({ funding, ...props }: TimelineItemData): JSX.Element {
+function TimelineItem({ funding, ...props }: TimelineItemData): JSX.Element | null {
   const { inTimezone } = useTimezone();
   const { data: currency } = useLocaleCurrency();
 
+  if (!currency) {
+    return null;
```
elliotcourant, line 87: Skeleton instead?

interface/src/components/goals/GoalItem.tsx
```diff
@@ -16,10 +16,14 @@ export interface GoalItemProps {
   spending: Spending;
 }
 
-export default function GoalItem({ spending }: GoalItemProps): JSX.Element {
+export default function GoalItem({ spending }: GoalItemProps): JSX.Element | null {
   const { data: locale } = useLocaleCurrency();
   const { data: fundingSchedule } = useFundingSchedule(spending.fundingScheduleId);
 
+  if (!locale) {
+    return null;
```
elliotcourant, line 23: skeleton,

Realistically all these components that would just render null instead is bad.

elliotcourant (commented):
There are a lot of additions of !null here from the robit.

Basically what I need to now is:

- Some hooks should have better types such that if error == nil then that means data is 100% present and we dont need to nil check
- This means adding error/loading state checks to pretty much everything that uses a query hook, OR using suspense instead.

- Also the types for the constructors are super fucked now since we are going from any to the type of the class. I'd like to come up with a partial helper that takes all of the keys of the class then makes their value the json type equivalent. For example, `Date` becomes string or RRule becomes `string` etc. This way I can have a type safe input for the constructor from a json response.

- There are clearly many places that are returning null or allow null when they should or shouldnt.

interface/src/components/FormTextField.tsx
```diff
@@ -22,21 +22,22 @@ export interface FormTextFieldProps extends InputProps {
...
-      return null;
+  const formikContext = useFormikContext<Record<string, any>>();
+  const getFormikError = (): string | undefined => {
+    if (!props?.name || !formikContext?.touched[props.name]) {
+      return undefined;
     }
 
-    return formikContext?.errors[props?.name];
+    // These renderers are keyed by a flat field name, so the formik error for that field is a plain string.
+    return formikContext?.errors[props.name] as string | undefined;
```
elliotcourant, line 40: God formik is going to be a mess to rip out

interface/src/components/MSelectFrequency.tsx
```diff
@@ -72,13 +73,13 @@ export default function MSelectFrequency(props: MSelectFrequencyProps): JSX.Elem
     <Select<string>
       {...props}
       disabled={props.disabled || formikContext.isSubmitting}
-      error={formikContext.errors[props.name]}
+      error={formikContext.errors[props.name] as string | undefined}
```
elliotcourant, line 76: How is this not implicit? If props.name isnt present in the errors map then it would be undefined? What does this casting even get us?

interface/src/components/MSelectFrequency.tsx
```diff
@@ -72,13 +73,13 @@ export default function MSelectFrequency(props: MSelectFrequencyProps): JSX.Elem
...
       label={props.label}
       name={props.name}
       onChange={onChange}
       options={options}
       placeholder={options.length > 0 ? 'Select a frequency...' : 'Select a date first...'}
-      value={value}
+      value={value ?? undefined}
```
elliotcourant, line 82: This is stupid? When would value be null such that it gets cast to undefined. Why is null not acceptable but undefined is?

interface/src/components/MSelectFunding.tsx
```diff
@@ -52,10 +52,10 @@ export default function MSelectFunding(props: MSelectFundingProps): JSX.Element
   }
 
   function createAndSetFunding() {
-    showNewFundingModal().then(result => formikContext.setFieldValue(props.name, result.fundingScheduleId));
+    showNewFundingModal().then(result => result && formikContext.setFieldValue(props.name, result.fundingScheduleId));
```
elliotcourant, line 55: When would result ever be falsy here? Is this because of if you close the funding modal without creating a new one then it does this?

interface/src/components/PullToRefresh.tsx
```diff
@@ -107,10 +107,10 @@ export default function PullToRefresh(): JSX.Element {
       if (document.querySelectorAll('[role="dialog"]').length > 0) {
         return;
       }
-      if (document.querySelector('ul')?.scrollTop > 0) {
+      if ((document.querySelector('ul')?.scrollTop ?? 0) > 0) {
         return;
       }
-      if (document.querySelector('form > div.overflow-y-auto')?.scrollTop > 0) {
```
elliotcourant, line 113: Since tailwind is gone this isnt even a thing anymore?

interface/src/components/PullToRefresh.tsx
```diff
@@ -107,10 +107,10 @@ export default function PullToRefresh(): JSX.Element {
       if (document.querySelectorAll('[role="dialog"]').length > 0) {
         return;
       }
-      if (document.querySelector('ul')?.scrollTop > 0) {
+      if ((document.querySelector('ul')?.scrollTop ?? 0) > 0) {
```
elliotcourant, line 110: I'm not sure this even matters since monetr now scrolls the entire body?

interface/src/components/Typography.tsx
```diff
@@ -67,7 +67,7 @@ export const textVariants = cva([styles.root], {
   },
 });
 
-type VariantProps = Omit<Parameters<typeof textVariants>[0], 'className' | 'class'>;
+type VariantProps = Omit<NonNullable<Parameters<typeof textVariants>[0]>, 'className' | 'class'>;
```
elliotcourant, line 70: What does this even do?

interface/src/components/ViewManager.tsx
```diff
@@ -25,7 +25,7 @@ function useViewContext<T extends string, Metadata extends {}, Form extends {}>(
   if (!context) {
     throw new Error('useViewContext must be used within a ViewManager');
   }
-  return context as ViewContextType<T, Metadata, Form>;
+  return context as unknown as ViewContextType<T, Metadata, Form>;
```
elliotcourant, line 28: Fuck you

interface/src/components/ViewManager.tsx
```diff
@@ -98,12 +98,16 @@ function ViewManager<T extends string, Metadata, Form>({
     [currentView, formData, metadata, updateFormData, updateMetadata, prevView, goToView, viewOrder, reset],
   );
 
+  // ViewContext holds its value with widened type arguments, so bridge our concrete value across to the provider.
+  // Consumers re-narrow it through useViewContext, which is why casting through unknown here is safe.
+  const contextValue = value as unknown as ViewContextType<string, unknown, unknown>;
```
elliotcourant, line 103: fuck you

interface/src/hooks/useAuthentication.ts
```diff
@@ -21,7 +21,7 @@ export interface Authentication {
 export function useAuthentication(): UseQueryResult<Authentication | undefined, unknown> {
   const select = useCallback(
     (data: Partial<Authentication>) => ({
-      user: Boolean(data?.user) && new User(data?.user),
+      user: (data?.user && new User(data.user)) ?? null,
```
elliotcourant, line 24: This was poorly typed to begin with, technically monetr relied on user being nullable in order to determine if the user was authenticated.

interface/src/hooks/useBankAccount.ts
```diff
@@ -26,7 +26,7 @@ export function useBankAccount(bankAccountId?: string): UseQueryResult<BankAccou
   return useQuery<Partial<BankAccount>, unknown, BankAccount | undefined>({
     queryKey: [`/api/bank_accounts/${bankAccountId}`],
     enabled: Boolean(bankAccountId), // Only request if we have a valid bank account ID to work with.
-    select: data => Boolean(data) && new BankAccount(data),
+    select: data => (data ? new BankAccount(data) : undefined),
```
elliotcourant, line 30: Realistically, when would data be falsy here? What was the situation that necessitated that?

interface/src/hooks/useIconSearch.ts
```diff
@@ -23,5 +23,5 @@ export function useIconSearch(name: string): Icon | null {
     retry: false,
   });
 
-  return data;
+  return data ?? null;
```
elliotcourant, line 26: Is this preventing returning undefined?

interface/src/hooks/useLocaleCurrency.ts
```diff
@@ -34,7 +34,10 @@ export default function useLocaleCurrency(forceCurrency?: string): UseQueryResul
   const locale = useCurrentLocale();
   const currency = useMemo(() => {
     // Return the first _defined_ currency.
-    return [forceCurrency, bankAccount?.data?.currency, me?.defaultCurrency, DefaultCurrency].find(value => !!value);
+    return (
+      [forceCurrency, bankAccount?.data?.currency, me?.defaultCurrency, DefaultCurrency].find(value => !!value) ??
+      DefaultCurrency
```
elliotcourant: I mean type wise the type system might just see `Array<string | undefined>` But is there a type safe way to tell the type system "yeah at least one of these is always defined" so that we dont need to do this dumb ?? at the end?

interface/src/hooks/useLunchFlowBankAccountsRefresh.ts
```diff
@@ -10,7 +10,7 @@ export default function useLunchFlowBankAccountsRefresh(): UseMutationResult<
   unknown
 > {
   return useMutation({
-    mutationFn: async (lunchFlowLinkId?: string): Promise<string> => {
+    mutationFn: async (lunchFlowLinkId: string): Promise<string> => {
```
elliotcourant, line 13: This is more correct actually yeah

interface/src/hooks/useLocaleCurrency.ts
elliotcourant:
```ts
type LastIsString  = [...(string | undefined)[], string];
```

This will do something close enough, where it promises that the last item in the array IS defined no matter what.

elliotcourant:
```ts
function fallback<T>(input: [...(T | undefined)[], T]): T {
  return input.find(value => Boolean(value))
}
```
This should be sufficient, though itd be neat to have a way to say "where T cannot be undefined"

elliotcourant:
```ts
function fallback<T extends NonNullable<T>>(input: [...(T | undefined)[], T]): T {
  return input.find(value => Boolean(value));
}
```

interface/src/hooks/useSignUp.ts
```diff
@@ -17,11 +17,7 @@ export interface SignUpResponse {
   requireVerification: boolean | null;
 }
 
-export interface SignUpError {
-  error: string;
-}
-
-export default function useSignUp(): (args: SignUpArguments) => Promise<SignUpResponse | SignUpError> {
+export default function useSignUp(): (args: SignUpArguments) => Promise<SignUpResponse> {
```
elliotcourant, line 20: I hate this, I wish promises had a way to specify the type of an error.

interface/src/modals/NewBankAccountModal.tsx
```diff
@@ -54,6 +54,12 @@ function NewBankAccountModal(): JSX.Element {
   };
 
   const submit = async (values: NewBankAccountValues, helper: FormikHelpers<NewBankAccountValues>): Promise<void> => {
+    // The locale is needed to convert the friendly balance into a stored amount, it should always be loaded by the time
+    // we can submit but bail just in case it is not ready yet.
+    if (!locale) {
+      return Promise.resolve();
+    }
```
elliotcourant, line 60: this will 100% cause bugs

docs/tsconfig.json
```diff
@@ -1,11 +1,12 @@
 {
   "compilerOptions": {
-    "baseUrl": ".",
     "target": "ESNEXT",
+    "types": ["node"],
     "lib": ["dom", "dom.iterable", "esnext"],
     "allowJs": true,
     "skipLibCheck": true,
     "strict": true,
+    "noUncheckedSideEffectImports": false,
```
elliotcourant, line 9: Need to figure out why this is necessary and how to remove it.

emails/tsconfig.json
```diff
@@ -1,19 +1,19 @@
...
   "compilerOptions": {
     "allowJs": true,
-    "baseUrl": "src",
     "esModuleInterop": true,
     "jsx": "react-jsx",
     "module": "ESNEXT",
-    "moduleResolution": "node",
-    "noImplicitAny": false,
+    "moduleResolution": "bundler",
+    "noUncheckedSideEffectImports": false,
```
elliotcourant, line 8: Ditto this?

emails/tsconfig.json
```diff
@@ -1,19 +1,19 @@
...
     "noImplicitReturns": true,
     "allowImportingTsExtensions": true,
     "sourceMap": true,
-    "strict": false,
+    "skipLibCheck": true,
```
elliotcourant, line 12: ?

interface/tsconfig.json
```diff
@@ -2,22 +2,24 @@
...
     "noImplicitReturns": true,
     "allowImportingTsExtensions": true,
     "outDir": "../server/ui/static",
     "sourceMap": true,
     "skipLibCheck": true,
-    "strict": false,
+    "strict": true,
+    // monetr's model classes hydrate through Object.assign(this, Partial<T>), which strictPropertyInitialization
+    // cannot see, so we leave just this one strict check off instead of asserting every model field with `!:`.
+    "strictPropertyInitialization": false,
```
elliotcourant: I want to remove this and update monetr's classes for client side objects to instead be strict about the fields.

interface/tsconfig.json
```diff
@@ -2,22 +2,24 @@
...
   "compilerOptions": {
     "allowJs": true,
-    "baseUrl": "src",
     "esModuleInterop": true,
     "jsx": "react-jsx",
     "module": "ESNEXT",
-    "moduleResolution": "node",
-    "noImplicitAny": false,
+    "moduleResolution": "bundler",
+    "noUncheckedSideEffectImports": false,
```
elliotcourant, line 9: also why this

interface/src/index.tsx
```diff
@@ -11,6 +11,9 @@ import '@monetr/interface/styles/styles.css';
 import '@monetr/interface/styles/index.scss';
 
 const container = document.getElementById('root');
+if (!container) {
+  throw new Error('failed to find the root element to mount monetr');
+}
```
elliotcourant, line 16: Might be better do just do `document.getElementById(...)!`

interface/src/env.d.ts
```diff
@@ -4,13 +4,11 @@ declare module '*.svg' {
...
 declare module '*.module.css' {
-  // biome-ignore lint/suspicious/noExplicitAny: This is just to make editors happy!
-  const classes: any;
+  const classes: { readonly [key: string]: string };
   export default classes;
 }
 declare module '*.module.scss' {
-  // biome-ignore lint/suspicious/noExplicitAny: This is just to make editors happy!
-  const classes: any;
+  const classes: { readonly [key: string]: string };
```
elliotcourant, line 11: Maybe time to look into https://rsbuild.rs/guide/styling/css-modules#type-generation

interface/src/util/sortAccounts.ts
```diff
@@ -3,7 +3,7 @@ import type BankAccount from '@monetr/interface/models/BankAccount';
 /**
  * sortAccounts will take an array of accounts and sort them by the account type and sub type priorities.
  */
-export default function sortAccounts(bankAccounts: Array<BankAccount> | undefined): Array<BankAccount> {
+export default function sortAccounts(bankAccounts: Array<BankAccount> | null | undefined): Array<BankAccount> {
```
elliotcourant, line 6: I dislike using both null and undefined. It should pretty much always be one or another unless its absolutely necessary to differentiate between the two.

interface/src/util/amounts.ts
```diff
@@ -63,7 +63,7 @@ export function getDecimalSeparator(locale: string): string {
   const numberWithDecimalSeparator = 1.1;
   return Intl.NumberFormat(localeAdjusted)
     .formatToParts(numberWithDecimalSeparator)
-    .find(part => part.type === 'decimal').value;
+    .find(part => part.type === 'decimal')?.value ?? '.';
```
elliotcourant: This the comma and the 2's should all be consts.

interface/src/testutils/queryText.ts
```diff
@@ -1,3 +1,3 @@
 export function queryText(query: string): string | null {
-  return document.querySelector(query).textContent;
+  return document.querySelector(query)?.textContent ?? null;
```
elliotcourant, line 2: I get it, the reason this is necessary is because this would otherwise be undefined, not null. So making it string | null means we have to cast it here. Maybe just update the type to be `string?` ?

interface/src/pages/transactions.tsx
```diff
@@ -165,7 +165,7 @@ function AddTransactionButton(): JSX.Element {
   );
 }
 
-function UploadButtonMaybe(): JSX.Element {
+function UploadButtonMaybe(): JSX.Element | null {
```
elliotcourant: Maybe jsx element isnt right here? Maybe its react.reactnode or whatever? I know there is a type that is nullable.

interface/src/pages/register.tsx
elliotcourant, line 33: Just make these | null

interface/src/pages/register.tsx
```diff
@@ -106,8 +106,8 @@ export default function Register(): JSX.Element {
     helpers.setSubmitting(true);
 
     return await signUp({
-      betaCode: values.betaCode,
-      captcha: values.captcha,
+      betaCode: values.betaCode ?? null,
+      captcha: values.captcha ?? null,
```
elliotcourant, line 110: Then you dont need silly ?? here?

interface/src/pages/login.tsx
```diff
@@ -55,7 +55,7 @@ export default function Login(): JSX.Element {
     helpers.setSubmitting(true);
 
     return login({
-      captcha: values.captcha,
+      captcha: values.captcha ?? undefined,
```
elliotcourant, line 58: There are so many goofy undefineds everywhere.

I'm going to add

https://biomejs.dev/linter/rules/no-useless-undefined/

https://biomejs.dev/linter/rules/no-useless-undefined-initialization/

https://biomejs.dev/linter/rules/use-nullish-coalescing/

interface/src/pages/verify/email/resend.tsx
```diff
@@ -71,7 +71,7 @@ export default function ResendVerificationPage(): JSX.Element {
   );
 
   const initialValues: ResendValues = {
-    email: emailFromQuery || undefined,
+    email: emailFromQuery || null,
```
elliotcourant, line 74: https://biomejs.dev/linter/rules/use-nullish-coalescing/

elliotcourant:
Oh and I definitely want `exactOptionalPropertyTypes`

PR #3337, opened 2026-06-19
fix: enable balance editing for manual bank accounts

### LINKED ISSUE #2311: feat(ui): Allow balances to be modified for manual accounts
Change the UI so that the Available, Current and Limit balances can be clicked on to be modified. Show a modal to edit the balance.

mvanhorn:
## Summary

- Uncomment and enable `availableBalance`, `currentBalance`, and `limitBalance` fields in the `putBankAccount` handler for manual (non-Plaid) accounts, including validation that `limitBalance` cannot be negative
- Extend `UpdateBankAccountRequest` in `useUpdateBankAccount.ts` to carry the three balance fields
- Add `FormAmountField` inputs for balances to the bank account settings page, gated on `link.getIsManual()`, so only manual accounts expose editable balances
- Add test cases: happy path (balance update round-trips through GET) and negative `limitBalance` rejection

## Why this matters

Manual bank accounts are managed entirely by the user, but the PUT endpoint had the balance fields commented out with TODO markers, making it impossible to correct balances after account creation. Fixes #2311.

## Testing

- PUT `/api/bank_accounts/:id` with `availableBalance`, `currentBalance`, and `limitBalance` on a manual account returns 200 with updated values reflected in the response and persisted on a subsequent GET
- PUT with `limitBalance: -1` returns 400 with `"Limit balance cannot be negative"`
- Plaid-linked accounts: balance fields sent in the request are ignored (existing behavior, handler branches on `PlaidBankAccountId == nil`)
- Settings UI shows balance fields only for manual accounts

### DIFF
--- a/interface/src/pages/bank/settings.tsx
+++ b/interface/src/pages/bank/settings.tsx
@@ -6,6 +6,7 @@ import { useLocation } from 'wouter';
 import type { ApiError } from '@monetr/interface/api/client';
 import { Button } from '@monetr/interface/components/Button';
 import Card from '@monetr/interface/components/Card';
+import FormAmountField from '@monetr/interface/components/FormAmountField';
 import FormTextField from '@monetr/interface/components/FormTextField';
 import MForm from '@monetr/interface/components/MForm';
 import MTopNavigation from '@monetr/interface/components/MTopNavigation';
@@ -23,6 +24,9 @@ import styles from './settings.module.scss';
 interface BankAccountValues {
   name: string;
   currency: string;
+  availableBalance: number;
+  currentBalance: number;
+  limitBalance: number;
 }
 
 export default function BankAccountSettingsPage(): React.JSX.Element | null {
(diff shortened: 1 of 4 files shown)


elliotcourant:
Thank you for your contribution! At a glance this looks good to me! Let me tinker a bit and I can see about getting this merged and into the next release

interface/src/pages/bank/settings.tsx
```diff
@@ -99,6 +108,9 @@ export default function BankAccountSettingsPage(): React.JSX.Element | null {
   const initialValues: BankAccountValues = {
     name: bankAccount.name,
     currency: bankAccount.currency,
+    availableBalance: bankAccount.availableBalance,
+    currentBalance: bankAccount.currentBalance,
+    limitBalance: bankAccount.limitBalance ?? 0,
```
elliotcourant, line 113: I might actually want to coalesce this to null, but the current controller doesn't do well handling it i dont think.

I have some work in progress for moving the controller entirely over to PATCH logic for bank accounts where passing null would actually remove the value versus having a 0 value, but that can come after this so no action needed on your part here.

elliotcourant (approved):
This looks good to me! I'll merge and send it out in the next release!

elliotcourant:
The tests that fail are failing because of PUT logic, this is a big reason I want to move to PATCH style endpoints. I can update the tests to resolve it

mvanhorn:
Appreciate the merge, @elliotcourant. Enabling balance editing for manual bank accounts fills an obvious gap.

mvanhorn:
Thank you @elliotcourant. Enabling balance editing for manual bank accounts is a sensible thing to let users do.

PR #3342, opened 2026-06-22
refactor(api): Upgrading from labstack echo v4 to v5

elliotcourant:


server/application/app.go
```diff
@@ -37,33 +39,40 @@ func NewApp(configuration config.Configuration, controllers ...Controller) *echo
...
+	if len(configuration.CORS.AllowedOrigins) > 0 {
+		app.Use(middleware.CORSWithConfig(middleware.CORSConfig{
+			AllowOrigins: configuration.CORS.AllowedOrigins,
+			AllowMethods: []string{
+				http.MethodDelete,
+				http.MethodGet,
+				http.MethodHead,
+				http.MethodOptions,
+				http.MethodPost,
+				http.MethodPut,
```
elliotcourant, line 59: Maybe this should be configurable too?

elliotcourant: I'll leave this alone, cors is basically not being used so this is fine as is for now. IF there if a need ill address it then

PR #3390, opened 2026-07-12
feat(api): Support API Key authentication

### LINKED ISSUE #2319: feat(api): Support some form of API key or long lived credential.
People want to use monetr's APIs and while they are not stable yet and are still changing, trying to hijack the cookie from the login endpoint is not very conducive to building something. It would be much nicer to have a dedicated API key and secret for each user<->account mapping.

- How are the keys generated
- How are they validated/cached? DB round trip for every request?
- How are they revoked/removed
- Should API key activity be tracked differently? Like a log tag indicating whether or not its a user or machine?

elliotcourant:
This introduces API key authentication support for monetr, API keys are
minted as ed25519 public and private keys. The server stores the public
key and the client is given the seed of the private key as a base32
secret. The secret is validated against the public key upon request
against the specific key ID it is for.

Some routes are not accessible via API key authentication, such as
billing related routes or authentication routes.

Resolves #2319


server/controller/api_key.go
```diff
@@ -0,0 +1,81 @@
...
+		key,
+		schemas.CreateApiKey,
+	)
+	if err != nil {
+		return err
+	}
+
+	if err := repo.CreateApiKey(c.getContext(ctx), key); err != nil {
+		return c.wrapPgError(ctx, err, "Failed to create API key")
+	}
```
elliotcourant, line 55: Need to add proof of work here

server/models/api_key.go
```diff
@@ -0,0 +1,97 @@
...
+	tableName string `pg:"api_keys"`
+
+	ApiKeyId      ID[ApiKey]  `json:"apiKeyId" pg:"api_key_id,notnull,pk"`
+	AccountId     ID[Account] `json:"-" pg:"account_id,notnull"`
+	Account       *Account    `json:"-" pg:"rel:has-one"`
+	Name          string      `json:"name" pg:"name,notnull"`
+	PublicKey     []byte      `json:"-" pg:"public_key,notnull"`
+	CreatedAt     time.Time   `json:"createdAt" pg:"created_at,notnull"`
+	CreatedBy     ID[User]    `json:"createdBy" pg:"created_by,notnull"`
+	CreatedByUser *User       `json:"-" pg:"rel:has-one,fk:created_by"`
```
elliotcourant, line 34: Might want to make this surface in the json so we can show info about "who" made the api key in the UI?

elliotcourant:
I think the frontend for this I'll add in another PR

